PR #5, opened 2017-03-20
Add comment about needed `cabal` packages

rauchg:
Compilation failed for me otherwise on a vanilla system with just `ghc`


b-studios:
For me it is enough to run `cabal install`. Maybe that was changed in the meantime. I am closing the PR, but thanks for your help!

PR #8, opened 2017-04-23
Changing jake and madoko to be local dependencies rather than global

Pauan:
This pull request changes `jake` and `madoko` to be local dependencies rather than global.

You can use `npm run` commands instead of `jake` commands.

----

Using global dependencies is bad because:

* It requires `sudo` permissions

* It prevents reproducible builds

* If the user already has `jake` or `madoko` installed with a different version, it will cause conflicts

* It pollutes the user's computer (deleting the `koka` folder isn't enough to uninstall Koka)

----

In addition to avoiding the above downsides, it also has some benefits:

* The user does not need to learn a separate build tool, they can just use `npm run`, which is the idiomatic way to use npm.

* It allows for swapping to a different build system in a backwards-compatible way, because the new build system can continue to use the exact same `npm run` commands.

* It improves integration with existing systems (such as continuous integration), because they automatically run the `npm test` command.

### DIFF
--- a/doc/spec/getstarted.kk.md
+++ b/doc/spec/getstarted.kk.md
@@ -49,19 +49,16 @@ Now we can build Koka itself:
 
        > npm install
 
-   If you are running on MacOSX or Unix, you may have to run this as
-   ``sudo npm install`` so that the ``npm`` package manager has enough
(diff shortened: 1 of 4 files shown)


Pauan:
@daanx :ping_pong:

PR #19, opened 2017-06-02
Adding in some more functions to array.kk

Pauan:


TimWhiting:
Going to close as out-of-date, so we can clean up the PR list. Thanks again!

PR #35, opened 2017-06-27
Adding in support for absolute file paths

### LINKED ISSUE #33: [BUG] Cannot compile absolute paths
There are two broken situations:

1. The file path breaks if it's absolute. This command:

   ```
   koka --compile --library --outdir="build" --target="js" "/absolute/path/to/koka/lib/toc.kk"
   ```

   Fails with this error:

   ```
   error: could not find: /absolute/path/to/koka/lib/toc.kk
   search path: <empty>
   ```

2. The `--include` flag breaks if it's absolute. This command:

   ```
   koka --include="/absolute/path/to/koka/lib" --compile --library --outdir="build" --target="js" "toc.kk"
   ```

   Fails with this error:

   ```
   error: could not find: toc.kk
   search path: /absolute/path/to/koka/lib
   ```

This is causing problems for Nix, because Nix stores everything in `/nix/store/`, so the Koka compiler can't find any files which are managed by Nix.

Pauan:
Fixes #33 

You can [add `?w=1` to the GitHub URL](https://github.com/koka-lang/koka/pull/35/files?w=1), this causes it to ignore whitespace which makes it easier to review the changes.

### DIFF
--- a/src/Common/File.hs
+++ b/src/Common/File.hs
@@ -9,7 +9,7 @@
     Internal errors and assertions.
 -}
 -----------------------------------------------------------------------------
-module Common.File( 
+module Common.File(
                   -- * System
                     getEnvPaths, getEnvVar
                   , searchPaths, searchPathsEx
@@ -28,7 +28,7 @@ module Common.File(
                   , isAbsolute
                   , commonPathPrefix
                   , normalizeWith
-                  , isLiteralDoc 
+                  , isLiteralDoc
 
                   -- * Files
                   , FileTime, fileTime0, maxFileTime, maxFileTimes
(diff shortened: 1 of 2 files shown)


Pauan:
This fixes some of the issues with absolute paths, but there is still a remaining issue:

When using `koka "/absolute/path/to/koka/lib/std/core.kk"` I get this error:

```
*** internal compiler error: Common.Name.qualify: Cannot use qualify on qualified names: (absolute/path/to/koka/lib/std/core,std/core/(<>))
```

I did some debugging, but it seems that there are a lot of places in the code which assume that paths aren't absolute.

I can try to fix all the places which assume non-absolute paths, but it will take some time, and it should probably be in a separate pull request. So I think this pull request should be accepted first.

TimWhiting:
The corresponding issue has been closed and Daan has made further changes recently to make absolute paths work better.  Let us know if you are still having issues. Thanks for the contribution!

PR #42, opened 2017-07-05
Adding in red black trees

Pauan:
This should allow for efficient implementations of immutable sorted sets and dictionaries.


Pauan:
I added in immutable sets in `std/data/set.kk`

Pauan:
I added in immutable sorted dictionaries in `std/data/sorted-dict.kk`

PR #53, opened 2017-07-14
Changing trace to accept any type

### LINKED ISSUE #20: trace function should be more useful
Currently, the `trace` function requires its argument to be a string. However, there are many things that you might want to log to the console which can't be (easily) converted to strings.

As such, I recommend changing the type of `trace` to `forall<a> (value: a) -> ()`

In addition, I suggest having the `trace` function automatically add in the filename, line number, and column number where the `trace` function is called.

So if you use `trace([1, 2, 3])` it will log something like this:

```
trace: [1, 2, 3]
  file: path/to/foo.kk, line 10, column 3
```

This makes it much easier to determine which trace output is which when there are multiple calls to `trace`.

There should probably be a second version of `trace` which accepts a message parameter, so if you use `trace("some message", [1, 2, 3])` it will log something like this:

```
trace: some message [1, 2, 3]
  file: path/to/foo.kk, line 10, column 3
```

Pauan:
Partial fix for #20

I haven't tested the C# code, but it probably works.

### DIFF
--- a/lib/std/core-inline.cs
+++ b/lib/std/core-inline.cs
@@ -380,10 +380,11 @@ public static class Primitive
   //---------------------------------------
   // Trace
   //---------------------------------------
-  public static void Trace( string msg )
+  public static void Trace( Object msg )
   {
-    System.Diagnostics.Debug.Print(msg);
-    Console.Error.WriteLine(msg);
+    string c = msg.ToString();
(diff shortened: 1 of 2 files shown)


TimWhiting:
With how trace is implemented now, I don't think this fix is applicable anymore, but thanks for the contribution!

PR #66, opened 2018-05-02
Find require.js in a saner way.

### LINKED ISSUE #4: "could not find requirejs" warning
When in `koka's` interpreter, typing anything in gives "could not find requirejs" warning/error (twice), after which the normal result is printed, e.g.
```
> 2
could not find requirejs
could not find requirejs
2

>
```

On the other hand, typing in shell:
```
$ node out/lib/interactive.js
2
$ nodejs out/lib/interactive.js
2
$
```
So it doesn't complain about not being able to find `requirejs`.

On my PC, `node` is an alias for `nodejs`, and `nodejs` version is v4.2.6. Also, my `$NODE_PATH` is `/usr/local/lib/node_modules:/usr/local/lib/node_modules/requirejs`.
Setting `NODE_PATH` to `koka's` dir doesn't help.

How to get the warning go away?

bgeron:
I took @dhil's patch (which improves path joining behavour); adding to that this pull request invokes `npm root` to find the correct search paths for `require.js`.

I think this fixes #4 .

On my machine, I have to locally `npm install requirejs` in order for the tests to work. I've not completed all tests yet (nor do I completely understand npm) but things look fine so far.


bgeron:
I had accidentally un-commented a trace command in my patch, which I've now re-commented and I've force-pushed.

dhil:
You may want to contact Daan via e-mail If you want this PR to be merged as he does not check here frequently.

bgeron:
Will do, thanks!

daanx:
Hi Daniel, Bram,
Thanks so much for looking into this! And I finally get my settings fixed with Github so I should finally get email whenever something happens. I want to create a more active community and will work on this :-)

I looked at the patches, but I think I found a more portable way to get the require behaviour (I think). I just checked in fixes into the `dev` branch. 
1. The handling of absolute paths under Unix should work now I think; I'll test it later under Ubuntu
2. The finding of requirejs and amdefine: when installing Koka, it is installed for Koka itself. I modified the compiler to get the amdefine and requirejs from there and copy it over to the output directory. `requirejs` is only needed when running in the browser. Actually, both will be deprecated once ES6 modules become more widespread...

Thanks again!

daanx:
Hi Bram,  (are you dutch?)


Thanks! I'll update the code to call `xdg-open` instead.


Nice you are using Koka -- always feel free to email me if you have ideas for new things or improvements.


Best,

-- Daan


________________________________

PR #68, opened 2018-06-09
Implicits

b-studios:
First draft of implementing implicits in terms of effect handlers

This PR is just for discussion and not meant to be merged


b-studios:
The current design allows users to write programs like

```scala
implicit x : int
implicit y : int
implicit z : bool

fun test2() : < ?x, ?y, ?z > int {
  if (?z) then ?x else ?y
}

val res3 = {
  implicit val x = 2
  implicit val y = 3
  implicit val z = True
  test2()
}
```

b-studios:
I now turned implicits into legal identifiers. Using the `?` prefix however conflicted with generated names. As a workaround I just picked the `^` prefix for now. I updated the above code example accordingly.

b-studios:
I added `with`-expression syntax since the implicit val bindings get quite clunky very fast:

```scala
with (^x = 4, ^y = ^x+2) test3()
```

for

```scala
{
val ^x = 4
val ^y = ^x + 2
test3()
}
```

b-studios:
Ok, I switched back to `?x` syntax. To avoid names that include `-?` I changed `makeHiddenName ` to replace the `?` by `implicit-` so operation name look like `.op-implicit-x`. Of course, there still can be name clashes if users define operations `implicit-x`.

PR #69, opened 2018-06-12
Feature/implicits

b-studios:
This PR encodes implicit parameters using algebraic effects.

PR for discussion. Replaces #68. See original PR for more discussion.


b-studios:
With first class implicit handlers we now can write code like

```scala
// forall<a,e>. (() -> <?x|e> a) -> e a
val x = implicit (?x = 3)
// forall<a,e>. (() -> <?x,?y|e> a) -> e a
val xy = implicit (?x = 3, ?y = 4)

effect reader<a> { fun read() : a }

// forall<a,e>. (() -> <?x,reader<int>|e> a) -> e a
val read42 = implicit (read = 42, ?x = read())

val res2 = {
  using implicit (?x = 3)
  ?x + ?x
}
```

b-studios:
With the new lazy implicit handler declarations, we can write code like:

```scala
implicit (?x = { println("asked for x"); 42 }) { ?x + ?x }
```
which will evaluate the body in braces every time the implicit argumentis being used. However, note that this is driven syntactically and thus is not equal to

```scala
fun getX() { println("asked for x"); 42  }
implicit (?x = getX) { ?x + ?x }
```
which will bind the function `getX` to the implicit argument.

In contrast

```scala
...
implicit (?x = getX()) { ?x + ?x }
```
will be eagerly evaluated and trigger the side effects once.

b-studios:
The latest changes now allow syntax like the following:

``` scala
effect dimensions {
  val width : int
  val height : int
}

implicit fun test(a: int): int

implicit val test2: int

val h = handler {
  fun test2() { 42 }
}

val h2 = handler {
  val test2 = 43
}
```

Still missing:

- [ ] value effect operations don't evaluate the value once, but are currently just desugared into tail resumptions
- [ ] there is no connection between value declarations, value definitions and value usage. The different syntaxes can just be mixed. We need to annotate the effect operation with a `value`-label and extend the typechecker to assert that value definitions (i.e. `val x = 42`) and value-usage (i.e. just `x` instead of `x()`) only can be used for declared value effects.

b-studios:
@daanx  The latest commit (e3f96ab) introduced the biggest breaking change: Dynamically bound functions are tail resumptive per default.

```js
handler {
  fun op() { /* ... no resume here ... */ }
}
```
To actually get access to the continuation we now need to prefix the function with `effect`:

```js
handler {
  effect fun op() { /* ... resume now available ... */ }
}
```

b-studios:
Using `?` as part of an identifier is mostly independent of this proposal, so I am gonna move these changes to a separate pull request.

PR #74, opened 2018-08-28
buildable with stack ghc

Nymphium:
Many people use Stack to manage several versions of GHC.
This PR  the haskellers use `stack ghc` to build Koka while others can use native `ghc` the same as it used to.
To build with stack, build with environment variable `build_with_stack`:
```shell-session
$ jake build_with_stack=true
```


b-studios:
Hey @Nymphium, thank you very much for your contribution! I haven't verified it on my local machine, but the CI builds look good. There are two things that come to my mind:

- Do both builds still work if we delete the generated file [Lexer.hs](https://github.com/koka-lang/koka/blob/master/src/Syntax/Lexer.hs)? It is generated from `Lexer.x` by using Alex.
- Maybe we should update the Readme to reflect that users can choose between the two build options?

Again: Thanks for allowing Koka to be built with Stack

Nymphium:
> Do both builds still work if we delete the generated file Lexer.hs? It is generated from Lexer.x by using Alex.

Same as using cabal, if installing alex by stack, it can generate Lexer.hs.
To ensure this, should I write something to CI removing Lexer.hs and generating it?

b-studios:
Yes, it would be really great if you could look into this. In another branch I also deleted the lexer file (since it is generated) but while it builds without problems on systems with alex properly installed [CI can't find alex](https://travis-ci.org/koka-lang/koka/builds/413871742).

Nymphium:
The reason of [this failure](https://travis-ci.org/koka-lang/koka/builds/413871742) is that `alex` was not on `$PATH`.
I added `~/.cabal/bin` to `$PATH` so that Lexer.hs is generated by `alex` with stack and even cabal, on  the CI.

.travis/setenv.sh
```diff
@@ -0,0 +1,9 @@
+if [[ "${build_with_stack}" ]]; then
+  mkdir -p ~/.local/bin
+  export PATH=$HOME/.local/bin:$PATH
+  travis_retry curl -L https://get.haskellstack.org/stable/linux-x86_64.tar.gz | tar xz --wildcards --strip-components=1 -C ~/.local/bin '*/stack'
+else
+  export PATH=$HOME/.cabal/bin:$PATH
+  nvm install node
+  nvm use node
```
b-studios: Is there a reason, why you only install `node` when building with cabal?

Nymphium: That's mistake. I'll fix it.

PR #75, opened 2018-09-20
Improve precision of resume analysis on let binders

b-studios:
In this PR I try to improve the precision of the resume analysis. In particular we want to have for

```
effect writer {
  fun write(n : int): ()
}
```

## Scoped
```
val ignore = handler {
  fun write(n) {
    val x = resume(())
    val y = ()
    println(n);
    resume(())
  }
}
```
and

```
val ignore = handler {
  fun write(n) {
    val x = resume(())
    val y = resume(())
    println(n);
  }
}
```

## Scoped Once
```
val ignore = handler {
  fun write(n) {
    val x = resume(())
    val y = ()
    println(n);
  }
}
```

## Tail
```
val ignore = handler {
  fun write(n) {
    val y = ()
    println(n);
    resume(())
  }
}
```

@daanx I am not sure whether I am missing something for recursive `defs`

### DIFF
--- a/src/Core/AnalysisResume.hs
+++ b/src/Core/AnalysisResume.hs
@@ -96,11 +96,20 @@ arExpr' appResume expr
       Let [DefNonRec def] expr
         | defName def == getName resumeName  -- TODO: too weak a check!! improve it!! we just need to skip the first definition..
         -> arExpr' appResume expr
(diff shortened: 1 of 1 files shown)


b-studios:
@daanx ping :)

PR #79, opened 2019-02-20
Fixing build for GHC 8.6.3

alexreinking:


daanx:
That's great. I see [here](https://prime.haskell.org/wiki/Libraries/Proposals/MonadFail) that we can also add a MonadFail instance but for just these cases this is fine :-)

PR #91, opened 2020-08-05
Fix build for GHC8.8

pzzp:


sternenseemann:
This is not an issue with the compiler version per se, but with the version of `base`. It'd be better to check the version of `base` in the ifdefs.

daanx:
Hi @pzzp, apologies for the long delay :-(  in the end I used #94 as it avoids the ifdefs -- hope you are not too discouraged; there was just so much changing a few months ago. Going forward the sources should be more stable.

PR #93, opened 2020-09-06
Make getInstallDir more robust in unusual cases

sternenseemann:
`getExecutablePath` is a more robust implementation of `getProgramPath`
which is shipped with `base`. It also should work on Windows (testing would
be appreciated!).

The main advantage is that `getExecutablePath` has not the issue that it
wouldn't come up with a path to the executable if it isn't in `PATH` or
invoked in an unusual manner (e. g. via a symlink). `getProgramPath` would
then sometimes just return `"koka"` (`getProgName` exactly) and even worse,
`dirname p` in `getInstallDi`r would be `""`. `getExecutablePath` guarantees in
such cases that we have an absolute path to work with.

---

See also <https://hackage.haskell.org/package/base-4.14.0.0/docs/System-Environment.html#v:getExecutablePath>


daanx:
Thanks so much and apologies for the late merge -- busy times with perceus :-)

PR #108, opened 2020-12-17
Propagate 'Nothing's from readline correctly

### LINKED ISSUE #25: Ctrl+D should quit the REPL
Right now, Ctrl+D displays the message `end of file` and does not quit the REPL. It is common practice for REPLs to quit when using Ctrl+D (at least on Linux).

However, it should only quit when the text prompt is empty. In other words:

```
> (Ctrl+D pressed)
```

In the above scenario, it should exit. However:

```
> some code or whatever (Ctrl+D pressed)
```

In the above scenario, it should *not* exit. Instead it should treat it as meaning "I am finished typing this expression, please parse/compile/execute it now".

fwcd:
## Fixes #25 

This lets users exit the interactive shell by pressing Ctrl+D.

### DIFF
--- a/src/Platform/cpp/Platform/ReadLine.hs
+++ b/src/Platform/cpp/Platform/ReadLine.hs
@@ -43,18 +43,19 @@ withReadLine historyFile io
 
 readLine roots prompt
   = do line <- readLines
-       return (Just line)
+       return line
   where
-    readLines :: IO String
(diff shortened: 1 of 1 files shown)


daanx:
Great -- thanks. This also fixes docker issues if one didn't start with a -i flag

PR #134, opened 2020-12-31
fix fedora and archlinux support

SchrodingerZhu:


daanx:
Awesome! Thanks so much :-)

PR #135, opened 2020-12-31
fix jakefile error

anqurvanillapy:
As title, to make CI happy.


daanx:
Thanks! However, we are deprecating the Jakefile -- is there a specific reason to use this?

anqurvanillapy:
> Thanks! However, we are deprecating the Jakefile -- is there a specific reason to use this?

I noticed that this PR https://github.com/koka-lang/koka/pull/134 somehow triggered the CI and caused this error... I'm not familiar with it actually :P

Issue #136, opened 2020-12-31
FreeBSD has no `fcopyfile`


SchrodingerZhu:
see https://github.com/freebsd/freebsd-src/blob/main/bin/cp/utils.c

daanx:
Thanks, and thanks for the PR; I will close this issue as we can discuss in the PR. :-)

PR #139, opened 2021-01-01
improve freebsd support

SchrodingerZhu:
- fix FreeBSD copyfile:
  it has no fcopyfile syscall so we need to mimic one from freebsd's cp
  - use `mmap` if filesize is less than `8MiB`
  - otherwise
    - create buffer (via `mi_malloc`)
    - try `copy_file_range` 
    - if `copy_file_range` failed, fallback to `read/write`
  
- detect self process via `KK_PROC_SELF "/proc/curproc/file"`
  In FreeBSD, this path can be possibly not mounted, but it is okay to fall back to generic way.

### DIFF
--- a/kklib/src/os.c
+++ b/kklib/src/os.c
@@ -154,6 +154,39 @@ static int os_copy_file(const char* from, const char* to, bool preserve_mtime) {
 #include <copyfile.h>
 #elif defined(__linux__)
 #include <sys/sendfile.h>
+#elif defined(__FreeBSD__)
+// from https://github.com/freebsd/freebsd-src/blob/master/bin/cp/utils.c
+#include <sys/mman.h>
+#include <sys/param.h>
+#include <limits.h>
+#include <errno.h>
+//Memory strategy threshold, in pages: if physmem is larger then this, use a 
+//large buffer.
+#define PHYSPAGES_THRESHOLD (32*1024)
+// Maximum buffer size in bytes - do not allow it to grow larger than this. 
+#define BUFSIZE_MAX (2*1024*1024)
+// Small (default) buffer size in bytes. It's inefficient for this to be
+// smaller than MAXPHYS.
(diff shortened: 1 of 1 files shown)


SchrodingerZhu:
this is not ready yet

SchrodingerZhu:
fixes #136

SchrodingerZhu:
@daanx I am not sure that whether `os_copyfile` requires to handle irregular files like links, symlinks, fifos. Currently only regular file copy is supported.

SchrodingerZhu:
to detect self path, one can also try /compat/linux/proc/

daanx:
Hi Yifan, thanks for the PR! However, I feel hesitant to merge this -- I think using `mmap` is just too much complexity/dependency for our runtime; also, using `read`/`write` is tricky as they can return `EAGAIN` or `EINTR` in which case they should be retried.  I looked at your commits and pushed a commit `bfe9ad08` and `0629f8a`  that should fix the issues on FreeBSD using your approach: on non Windows/macOS/Linux, try `copy_file_range` first, and then fall back to a generic copy method with a buffer. 
Can you try if it works on FreeBSD?   Thanks again

SchrodingerZhu:
@daanx hi! I think your commits is much more robust and clearer. Tested and happy to see it fixed.

BTW, there is an additional problem:

https://www.freebsd.org/cgi/man.cgi?query=strerror_r&apropos=0&sektion=3&manpath=FreeBSD+7-current&format=html

```
out/v2.0.15/clang-release/std_core.c:280:13: warning: incompatible integer to pointer conversion initializing 'char *' with an expression of type 'int' [-Wint-conversion]
      char* serr = strerror_r(err, buf, 255); buf[255] = 0;
```

daanx:
Thanks Yifan -- I think the `strerror_r` is fixed with the latest `dev`  :-)

SchrodingerZhu:
fixed!

PR #145, opened 2021-01-09
Resolve VS Code extension not starting

ToucheSir:
Removed some line items in the extension configuration that were preventing extension startup on a recent version of VS Code (1.52.1).

- reference to non-existent main file
- `onLanguage:koka` depends on the contributed language config, but that doesn't seem to load unless the extension is activated? I'm not sure whether this chicken-and-egg behaviour is intentional on VS Code's part, but removing the activation event breaks the cycle.


daanx:
Great! Thanks so much @ToucheSir

PR #161, opened 2021-03-10
Add emacs support

kamoii:
Add a `koka-mode` major mode for emacs.
Syntax hightlights comments, strings, keywords and function name. 
Far from complete, but I think enough for a starting point.


daanx:
Wow -- that is awesome @kamoii !   Thanks so much. I'll look into if it is easy to install it by default (just like we do for vscode and atom)

PR #167, opened 2021-03-24
Update samples to use new append and add missing samples in all.kk

siraben:


daanx:
Thanks so much @siraben

PR #170, opened 2021-04-04
Support pattern matches in parameters

stevenfontanella:
* Support pattern matches in parameter lists
* Add Show instances for Program/Expr types
* Add more Trace functions
* Fix a unicode test and compiler error in lexer.lex


daanx:
Thanks Steven! Nice and clean code :-)  
Let's discuss further on Skype soon; some items to discuss:

- currently the new grammar for "parameters" allows pattern matches also in external definitions and operation declarations but I think we need to limit it to anonymous lambdas only (perhaps in named functions as well?)
- we can use hidden names for the pattern bindings (by prepending a .)
- I am a bit concerned with the extra dependency on auto derived Show instances but perhaps I should become more modern :-)
- wildcard pattern names

Thanks again

doc/spec/grammar/parser.y
```diff
@@ -531,6 +518,31 @@ argument    : expr
             | identifier '=' expr                  /* named arguments */
             ;
 
+/* parameters: separated by comma */
+
+parameters  : parameters1
```
stevenfontanella, line 523: The `operation` production still uses this production (instead of `normalparams`), but it still rejects patterns in operation declarations, e.g. this is rejected:

```haskell
    with handler {
        fun yield(Cons(x,xs) : list<int>) { a.println }
    }
```

This is the behavior we want, but why is it working this way? Should I change `operation` to use `normalparams`?

daanx:
Thanks so much @stevenfontanella

PR #180, opened 2021-07-17
Fix escape and package name for Arch Linux.

withoutK:


daanx:
Thanks so much! I integrated the fixes by hand directly into the `dev` branch as there were many whitespace changes (and conflicts with master). Best, Daan

PR #181, opened 2021-07-24
Allow constructors to be first-class values

camelid:
Before this change, using a constructor as a first-class value would
trigger a C compiler error. For example, `val x = Just` would cause the
C compiler to return a "not enough arguments" error.

After this change, constructors are correctly treated as first-class
values. This is accomplished by creating a lambda "on the fly" in the
C backend that looks like `fn(x, xs) { Cons(x, xs) }`.

Here is a longer example of code that used to not work, but does now:

    fun main() {
        val list_of_maybes = map([1, 2, 3], Just)
        val list_of_ints = map(
            list_of_maybes,
            fn(maybe_el) { match(maybe_el) { Just(el) -> el } }
        )
        println(show(list_of_ints)) // prints [1,2,3]
    }

### DIFF
--- a/src/Backend/C/FromCore.hs
+++ b/src/Backend/C/FromCore.hs
@@ -503,7 +503,7 @@ genTypeDefPost (Data info isExtend)
 
        -- generate functions for the datatype (box/unbox)
        genBoxUnbox name info dataRepr
-  where    
+  where
     ppStructConField con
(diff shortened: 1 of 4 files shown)


daanx:
Hi Noah, great to see you here  -- and apologies for the long delay for this PR! I should have been more pro active :-(  I am closing the PR as new simplifier/inliner already fixes this issue. Apologies again, feel free to contact me if you are looking for interesting projects. Best, Daan.

PR #193, opened 2021-09-15
Koka support for (neo)vim

neunenak:
This commit adds the code for a basic (neo)vim Koka plugin to the support/ directory.


daanx:
Thank so much!

Issue #212, opened 2021-10-27
ddouble1 and ddouble2 test failures


philderbeast:
There are two failures reported with `stack test --fast`:

```
> git status 
On branch dev
Your branch is up to date with 'origin/dev'.

nothing to commit, working tree clean
> git log
commit 171d163c1eec77223b1c86a578ca08af5bd906dd (HEAD -> dev, origin/dev)
Author: Daan Leijen <daan@microsoft.com>
Date:   Mon Oct 25 08:49:12 2021 -0700

    fix attribute on vector_buf_borrow

> stack test --fast
...
Failures:
  1) lib ddouble1
...
  To rerun use: --match "/lib/ddouble1/"
...
  2) lib ddouble2
...
  To rerun use: --match "/lib/ddouble2/"

Randomized with seed 1796457736

Finished in 142.2684 seconds
264 examples, 2 failures

koka> Test suite koka-test failed
Completed 2 action(s).
Test suite failure for package koka-2.3.3
    koka-test:  exited with: ExitFailure 1
Logs printed to console

daanx:
Hi @philderbeast ,  apologies for the late reply and thanks for trying Koka :-) 
Yes, the `ddouble` tests tend to fail as different systems have slightly different (rounding) results for some double math. With the 128-bit `ddouble` having 31 digits of precision those differences show up :-) ..  we need to improve our test suite I to be more lenient in the required digits of precision.  The real goal would be to be fully deterministic across systems though as that is one of the guarantees that Koka should bring when being in the `pure` effect! So, I am still hoping to fix the particular points where the math libraries differ across systems and make Koka deterministic for them. There was a cool paper at the last PLDI about something similar except I think it was for 32-bit floats though (?) not sure now.

TimWhiting:
I haven't seen this on MacOS for some time now. Was this fixed?

PR #213, opened 2021-10-28
Add github actions.

philderbeast:
Adds github actions for cabal and stack (lts-18.14) for ghc-8.10.7. Removed `regex-compat-tdfa` and `json` from extra-deps as they're included with lts-18.14. Opted not to mess with `./stack.yaml` at lts-18.8 for ghc-8.10.4 as it has commentary I didn't want to mess with. The stack runner yaml configuration uses `./stack/stack-8.10.7.yaml`.

I've seen test failures but that is a separate issue https://github.com/koka-lang/koka/issues/212.

The ci runners are triggered on: 

```yaml
on:
  push:
    branches: [ci-cabal, add/github-actions]
  pull_request:
    branches: [master]
```

Later we could remove `add/github-actions` from the push branches list but for now it is useful for triggering both stack and cabal runners.


TimWhiting:
We have a few github actions now, thanks for the contribution!

PR #215, opened 2021-10-28
Make tooltip overflow and use SCSS

### LINKED ISSUE #214: Better tooltip
1. tooltip showed out of window
2. tooltip clipped by code <div>

Problem 1 can be fixed by using `title` attribute like `<div title="hello">` since the tooltip is not clickable anyway. If styling is needed, we should use Javascript to make sure the tooltip stays on the page.

locriacyber:
SCSS allows cleaner code and line comments.
I'm not sure what `xfont-size` and `xwhite-space` are so I commented them out.

This also fixes #214 partially by allowing tooltips to display outside the code DOM element.


daanx:
Thanks so much for the PR @locriacyber ! however, as I discuss in #214 I hope to avoid SCSS for now. Sorry for that!

PR #221, opened 2021-12-08
Fix incorrect reference count when applying a bigint modulo to a negative smallint

rubikscraft:
Right now this code replicates the crash:
```koka
struct testing
  p: int

fun main(){
  val a = Testing(1000000000000000000)
  println(a.p) // 1000000000000000000
  val b = -1 % a.p
  println(b) // 999999999999999999
  println(a.p) // empty line
  val crash = 1 % a.p // Segfault
  println("bye") // Is never run because crash
}
```

a.p is incorrectly freed by the modulo operation, this means you can now still use the freed value after the modulo. This ofcourse produces incorrect results and causes a crash.

After some debugging I've managed to locate the problem to the `kk_interger_sub` and `kk_integer_add` functions. These decrement the rc of their parameters, so when these are not incremented beforehand they will get freed. This is what happens to the right paramter of the modulo in this case.

So I have fixed the bug by incrementing the rc before calling these functions. I'm not exactly sure if this is the right place in the code to increment them though, so feel free to modify if necessary.

### DIFF
--- a/kklib/src/integer.c
+++ b/kklib/src/integer.c
@@ -1401,7 +1401,7 @@ kk_integer_t kk_integer_cmod_generic(kk_integer_t x, kk_integer_t y, kk_context_
 // - x `mod` 2^n == and(x,2^(n-1))  for any x, n
 // - Euclidean division behaves identical to truncated division for positive dividends.
 kk_integer_t kk_integer_div_mod_generic(kk_integer_t x, kk_integer_t y, kk_integer_t* mod, kk_context_t* ctx) {
-if (kk_integer_is_zero_borrow(y)) {
+  if (kk_integer_is_zero_borrow(y)) {
     // div by zero
     if (mod!=NULL) {
       *mod = x;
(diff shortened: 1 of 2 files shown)


anfelor:
Oh, great catch! Yes, I think we have to add the dups there. We could also put the later `kk_integer_drop(y,ctx);` into an if-statement:
```
if(!kk_integer_is_neg_borrow(m) || mod == NULL) {
  kk_integer_drop(y,ctx);
}
```
This would enable more reuse in the calls to `kk_integer_add`/`kk_integer_sub` as they try to reuse the larger of their arguments. As `m` is the modulo wrt `y`, `y` is always larger than `m`, but we can only reuse `y` if we don't keep its reference count above one by dupping it at the function call.

rubikscraft:
Yeah that would indeed be better, so I changed the code accordingly. I did however need to make some changes, because the `m` variable is changed by the `kk_integer_sub` and `kk_integer_add` functions. So it would produce a different result when `kk_integer_is_neg_borrow(m)` is called by the new if statement.
I fixed it by storing the result of the first `kk_integer_is_neg_borrow(m)` in a bool. This makes sure that both if statements use the same value.

daanx:
Ha, thanks @rubikscraft  :-)   and thanks @anfelor  for your kind reply. 
I actually like the original fix best because the updated one is tricky with the if condition depending on the exact control flow befor it; a dup/drop on an integer is _fast_ so no problem sometimes doing a dup/drop pair to many if that makes the code more clear. Can you update once more to the original and I'll merge it in.

rubikscraft:
@daanx I reverted it back to the first solution.

daanx:
:-)

PR #223, opened 2021-12-16
Clarify doc for with-handlers

dabrahams:
As written, there was not enough context for me to understand the section, reading the tour from the top.  In describing the desugaring, it seemed to be explaining `with handler` in terms of `handler`, for which there was no explanation.  I couldn't make head or tail of the section until got to the very last sentence.

I took the liberty of changing the `emit-console` names to `hello-console`, which IMO better describes the semantics and will help to guide the reader along; I hope you don't mind!  Also added a missing antecedent for “this,” a habit an editor once drilled into my brain.

### DIFF
--- a/doc/spec/tour.kk.md
+++ b/doc/spec/tour.kk.md
@@ -297,20 +297,26 @@ it is all just function applications with minimal syntactic sugar.
 #### With Handlers  { #sec-with-handlers; }
 
 The `with` statement is especially useful in combination with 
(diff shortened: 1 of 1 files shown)


daanx:
Awesome -- thanks for the PR. It is hard to write the "book"; there are both new abstract concepts (like effect handlers, effect types) and a new language with syntax inventions :-).

PR #257, opened 2022-04-11
Fix the documentation for unsafe-nostate

iacore:
Also added `unsafe-no-st` for consistent API.


TimWhiting:
This has been fixed with recent refactorings of the core libraries. Thanks for the contribution!

PR #266, opened 2022-05-09
Update concat operator

jO-Osko:
Some of the comments (and excluded tests) used the old version for concatenation `+` instead of `++`


burjui:
It seems that Koka would benefit from something like doc tests that Rust has.

TimWhiting:
Sorry for the major refactor / splitting of the core that made this PR rather out of date with conflicts. Would you like to update the PR, or would you like me to just go ahead and fix these?

TimWhiting:
I've made the fixes in the current std libraries. The trace statements weren't being used so this has no functional difference, but makes it less confusing to new contributors. Thanks!

PR #269, opened 2022-05-15
Fix JS backend

pashutk:
There is probably a typo in _max_uint32 (and a couple of other) variable names as this module references `_max_uint32n` but doesn't declare `_max_uint32`. Also, the JS code isn't valid since declaring two consts with the same name isn't allowed in js.

Screenshot 2022-05-15 at 23 04 53


TimWhiting:
I believe this has been resolved on the latest dev. Thanks for the contribution!

PR #284, opened 2022-06-25
mod: [doc] performance -> inefficiency

mitsuji:
Hi. I'm big fan of Koka and Perceus.
Thank you for creating a great programming language.

I read the document and found something that might be corrected,
So I will send a PR.

There are expressions in the documentation
as if CG languages are a more efficient programming language than C/C++/Rust.

Especially the "our goal is to be within 2x of the performance of C/C++"
part seemed to be understood incorrectly.

As a experiment, I replaced "performance" with "inefficiency".
If there is a better way to fix it, that is fine.

Thank you.


konsumlamm:
What? "Inefficiency" means basically the opposite of "performance", did you mean "efficiency"? What about "performance" can be understood incorrectly?

Ugzuzg:
Inefficiency is probably not the best word, but I agree with it possibly being confusing. 2x performance from the charts could be read as "2 times more performant".

PR #292, opened 2022-07-30
Document map type as generic over effect

roetlich:
Sorry for making such a short PR, but I  got confused by the doc string of map. It says map can only be used with total functions, but the type signature seems to indicate otherwise. 

The guide seems to also contradict the doc string: https://koka-lang.github.io/koka/doc/book.html#why-effects


TimWhiting:
This looks like it was fixed in recent updates, thanks for the contribution!

Issue #103, opened 2020-12-09
Reproduced parsing segfault


patrickmn:
Ran into this interesting issue

```bash
wget https://gist.githubusercontent.com/patrickmn/32b279b6188a3239b13cd6bfa4df0008/raw/a95cecdfc47601399dafcc52d10fefe71d7726b7/day02.kk
wget https://gist.githubusercontent.com/patrickmn/32b279b6188a3239b13cd6bfa4df0008/raw/9e588e0488881e00ee060e1a95002778d13ff3ee/day02.txt
koka -e day02.kk # fails after 756 parses
                 # replace `whitespace0()` with `char('\n')` and it fails after 800 parses
# ...
# 754
# 755
# 756
# Segmentation fault (core dumped)
# failure while running program:
#  "out/v2.0.12/gcc-debug/day02"

koka -O1 -e day02.kk # doesn't fail
koka -O2 -e day02.kk # doesn't fail
```

```koka
module day02

import std/os/file
import std/os/path
import std/text/parse

struct entry {
  min : int
  max : int
  character : char
  password : list<char>
}

fun parse(s : string) : parse-error<list<entry>> {
  with s.slice.parse-eof
  var n := 0
  many {
    val from = pint()
    char('-')
    val to = pint()
    char(' ')
    val chr = alpha()
    char(':')
    char(' ')
    val chars = many(alpha)
    whitespace0()
    n := n + 1
    trace(n.show)
    Entry(from, to, chr, chars)
  }
}

fun main() {
  match ("day02.txt".path.read-text-file.parse) {
    ParseError(msg, _rest) -> println("parsing failed: " + msg)
    ParseOk(_parsed, _rest) -> println("parsed")
  }
}
```

Koka 2.0.12, 15:19:32 Dec 2 2020 (ghc release version)

gcc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0

https://gist.github.com/patrickmn/32b279b6188a3239b13cd6bfa4df0008

daanx:
Ouch! This is probably due to running out of stack space (on  linux, use `ulimit -s unlimited` to work around this). 
If not, it is a real bug but since it works in O2 it looks like a tail-call optimization (or lack of it)

The current parser is also not super -- it was constructed hastily to remove the dependency on regex but I would love to see it improved; in particular to be more eager (like Parsec, with an explicit `try`). As such, it may be that the parser is using more and more stack space in certain situations (not sure)

[...]

PR #317, opened 2022-12-09
Start of Koka self-hosted implementation

TimWhiting:
Working in this commit:
- Lexer (I'm looking into optimizing this by improving the parser machinery).
- Start of map implementation in Koka's std library. (Currently based on Haskell's Map, but I might switch to Red-Black trees with a FBIP implementation)
- Pretty printer
- Name Utilities

The files should all compile without errors

I've verified the Lexer by lexing the lexer.kk itself.
The lexer is rather slow right now, but I've started looking into the parser machinery as mentioned here: https://github.com/koka-lang/koka/issues/103#issuecomment-742852707

Before I go any farther I'd like to seek input from @daanx.

I think it would be great to have a self-hosted compiler for the following reasons:
- Make it easier for users to contribute fixes
- Demonstrate the usefulness of effects & fbip in a large-scale application
- I prefer Koka syntax

I also understand that Koka is currently a research language and that this introduces a larger maintenance burden during a transition period from the Haskell compiler, and the fact that Haskell has more libraries that can be used for the vscode extension protocol, and other helpful utilities like the Alex lexer library and Parsec, which I have definitely felt the lack of when trying to make the Lexer fast.

Regardless of whether this would eventually become the default compiler, I have found it a useful project to get familiar with Koka & Haskell & Koka's Haskell compiler, so I'll probably continue work on it for those reasons, and would welcome contributions from others who are also interested in this.

### DIFF
--- a/lib/std/core.kk
+++ b/lib/std/core.kk
@@ -414,12 +414,12 @@ pub fun append(xs : list<a>, ys : list<a> ) : list<a>
 
 // Fold a list from the right, i.e. `foldr([1,2],0,(+)) == 1+(2+0)`
 // Note, `foldr` is less efficient than `foldl` as it reverses the list first.
-pub fun foldr(xs,z,f)
-  xs.reverse.foldl(z) fn(x,y) f(y,x)  
+pub fun foldr(xs : list<a>, z : b, f : (b, a) -> e b) : e b
+  xs.reverse.foldl(z) fn(x,y) f(x,y)
 
 // Fold a list from the left, i.e. `foldl([1,2],0,(+)) == (0+1)+2`
 // Since `foldl` is tail recursive, it is preferred over `foldr` when using an associative function `f`
-pub fun foldl(xs,z,f)
+pub fun foldl(xs : list<a>, z : b, f : (b, a) -> e b) : e b
   match xs
     Cons(x,xx) -> foldl(xx,f(z,x),f)
     Nil        -> z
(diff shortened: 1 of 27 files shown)


TimWhiting:
Marking as draft for now, but willing to incrementally add the self-hosted implementation in chunks if that is better.

timjs:
Very brave to start working on this! Maybe you can split this up in a PR for an extended and more structured core lib first. Would be great to have general access to a map implementation for example. Next to an LSP implementation, it is something I'm missing atm. Building a community and great tools for Kok's would be awesome!

timjs:
About an optimised lexer/parser. Don't know if it's helpful, bu I experimented with an modular interface around parsers a la MegaParsec but stumbled on some internal compiler errors. See #258.

TimWhiting:
> About an optimised lexer/parser. Don't know if it's helpful, bu I experimented with an modular interface around parsers a la MegaParsec but stumbled on some internal compiler errors. See https://github.com/koka-lang/koka/discussions/258.

@timjs 
I'll definitely take a look!

> Maybe you can split this up in a PR for an extended and more structured core lib first. Would be great to have general access to a map implementation for example.

Good thought. There is some considerations for the map interface though, and collections in general.

1. Haskell uses typeclasses to enforce that you can sort the keys. Koka does not have typeclasses yet.
2. The sorting function could be passed into every method (inconvenient)
3. The sorting function could be an effect - inconvenient when you have two maps with different key types because of the swap rule for generic effect types (https://github.com/koka-lang/koka/discussions/247)
4. The sorting function has to be stored in the map type (probably not on every node, but as a wrapper object). The wrapper object is probably not a big deal if just using the map interface, but could be confusing if you want to traverse the map yourself with `match`

I've actually put some thought and research into using generic effect types to implement a typeclass-like constraints on the types of arguments of functions, by getting rid of some of the pain points of approach 3. I've got some basic stuff working such as automatically inserting user-defined default handlers for types. This is something I'm considering exploring further as a part of my PhD, but I'm curious if it aligns with what @daanx was thinking about for typeclasses in Koka, since he said he had a version of them in a design phase. I've got a branch for this and the start of a research proposal, and can discuss it more if there is interest. Mostly it involves default effects, and type-indexed generic effect handlers. I'm interested in pursuing this for my PhD as well as some optimizations of effect handlers such that the linear property could be inferred or proved via static analysis.

TimWhiting:
My work on this has moved elsewhere

PR #321, opened 2023-01-08
List additions

Matthew-Mosior:
This adds the permutations function which is a port of the OCaml permutations function found in this Rosetta Code page:

https://rosettacode.org/wiki/Permutations#OCaml

### DIFF
--- a/lib/std/core.kk
+++ b/lib/std/core.kk
@@ -713,6 +713,67 @@ pub fun permutations ( l : list<char>
     1 -> [l]
     _ -> aux(0,l.vector)
 
+// Generate all permutations.
+pub fun permutations ( l : list<int>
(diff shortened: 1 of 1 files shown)


Matthew-Mosior:
@daanx  Its odd, using the list<int> permutation implementation, there seems to be an issue when using a list > 8 in length (i.e. [1,2,3,4,5,6,7,8,9]), I'm seeing:

`core dumped.` on my ubuntu linux workstation

and 

`rosetta error: unexpectedly need to EmulateForward on a synchronous expection` on my M1 mac running in rosetta (emulating x86_64 environment).

It similarly crashes on my M1 mac when run without rosetta mode.

But testing by inputting a list of length =< 8 works fine.

PR #323, opened 2023-01-31
Fix 2

iacore:
I was trying to compile this [example project](https://gist.github.com/locriacyber/1d4cd15f4cadb8a24b3793e80d3a5b15), but `kk_int8_unbox` was missing. So I added it

I simply changed 16 to 8 from the code for int16.


mtoohey31:
Appears to be superseded by #388.

Issue #322, opened 2023-01-29
The executable Koka complied doesn't get permission on arm-macOS


LittleJianCH:
For example, this is the example on the Koka website:

``` koka
// A generator effect with one operation
effect yield<a>
  fun yield( x : a ) : ()

// Traverse a list and yield the elements
fun traverse( xs : list<a> ) : yield<a> () 
  match xs
    Cons(x,xx) -> { yield(x); traverse(xx) }
    Nil        -> ()

fun main() : console () 
  with fun yield(i : int)
    println("yielded " ++ i.show)   
  [1,2,3].traverse
```

Store it in `main.kk` and run the command `koka main.kk -o main && ./main`, you will get the message below:

``` shell
compile: main.kk
loading: std/core
loading: std/core/types
loading: std/core/hnd
check  : main
linking: main
created: .koka/v2.4.0/clang-debug/main
created: main
zsh: permission denied: ./main
```

To run the executable, you have give the executable permission by running `chmod +x ./main`. I think this is a part of work the compiler should do.

My platform is M1 Pro MacBook.

tomsib2001:
Same issue on Koka 2.4.2, Ubuntu 20.04

JanBeh:
Same with Koka 2.4.2 on FreeBSD 13.2

TimWhiting:
Is this still an issue? What permissions does the file have?

LittleJianCH:
I directly applied chmod -x to the compiled files in this pr(#324). It works well on my machine.

TimWhiting:
Is this still an issue? I have not seen this on my arm-macOS.

Dessertion:
This is still an issue on Koka 3.1.2, on openSUSE Tumbleweed.

ryota2357:
This is still an issue on Koka 3.2.2, on my aarch64-darwin (macOS 15.5)

I installed koka using Nix with devShell:

<details>
<summary>koka.nix</summary>

```nix
{
  lib,
  stdenv,
  fetchFromGitHub,
  pkgsHostTarget,
  haskellPackages,
  cmake,
  makeWrapper,
}:

let
  version = "3.2.2";

  src = fetchFromGitHub {
    owner = "koka-lang";
    repo = "koka";
    rev = "v${version}";
    hash = "sha256-k1N085NoAlxewAhg5UDMo7IUf2A6gCTc9k5MWMbU0d0=";
    fetchSubmodules = true;
  };

  kklib = stdenv.mkDerivation {
[...]

PR #324, opened 2023-02-01
Give final exe the executable permission

LittleJianCH:
More information about this pr can be found in #322.

I don't have any Linux or Windows machine handy. This pr was only tested on arm-macOS.

I'm also not sure if this bug is only appeared on macOS.


daanx:
Thank you for bringing this up -- we switched to using the home directory instead of the temporary directory to avoid the issue altogether.

PR #325, opened 2023-02-11
Improve effect type annotations

TimWhiting:
- Adds equality & substring for slices.
- Adds missing types
- Adds missing effect polymorphism

### DIFF
--- a/lib/std/core.kk
+++ b/lib/std/core.kk
@@ -339,9 +339,27 @@ pub fun unzip( xs : list<(a,b)> ) : (list<a>,list<b>)
   // todo: implement TRMC for multiple results
   fun iter( ys, acc1, acc2 )
     match ys
-      Cons((x,y),xx) -> iter(xx,Cons(x,acc1),Cons(y,acc2))
-      Nil            -> (reverse(acc1),reverse(acc2))
-  iter(xs,[],[])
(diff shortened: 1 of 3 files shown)


lib/std/core.kk
```diff
@@ -414,12 +414,12 @@ pub fun append(xs : list<a>, ys : list<a> ) : list<a>
 
 // Fold a list from the right, i.e. `foldr([1,2],0,(+)) == 1+(2+0)`
 // Note, `foldr` is less efficient than `foldl` as it reverses the list first.
-pub fun foldr(xs,z,f)
-  xs.reverse.foldl(z) fn(x,y) f(y,x)  
+pub fun foldr(xs : list<a>, z : b, f : (b, a) -> e b) : e b
```
xialvjun, line 417: It seems we needn't to write the type annotations because `xs.reverse` mean `reverse(xs)`. And since there is `pub fun reverse(xs : list<a>) : list<a>`, `xs` just has type `list<a>`. `z and f` are the same, they all should can be type inferenced.  But I didn't really write koka, I just have read the koka book, and I'm watch koka. If I'm wrong, ignore me.

TimWhiting: Yeah, the annotations aren't necessary, but it makes it clear to the reader of the code the intention. (Pretty much all of the rest of the std library has annotations).

TimWhiting:
@daanx 
Can you take a look at this?

daanx:
I'll merge it and fix partition but will comment out `substr` for now. Thanks!

PR #327, opened 2023-02-22
allow generated to be used by Deno

iacore:
Previously Deno was detected as Node


daanx:
Thank you! I added the fix to the latest `dev` branch.

PR #330, opened 2023-03-10
Add cconsole.h to koka.cabal

Lysxia:
Without it, `cabal install` fails (while `cabal build` succeeds either way).


TimWhiting:
Thanks for the contribution! `koka.cabal` actually gets generated from `package.yaml` so I've changed it there as well to prevent it from being removed the next time `koka.cabal` get's generated.

PR #331, opened 2023-03-16
fix slice advance function and test with regex

TimWhiting:
While using regular expressions for parsing I ran into some really weird behavior. After quite a bit of debugging I found the issue in the standard library's advance function for slices. Here is the fix, along with a host of tests for both negative / positive advances.


daanx:
Awesome!

PR #335, opened 2023-04-09
Generate (==) & show

TimWhiting:
Generate .show / (==) methods for user datatypes. One issue is that compile time type errors are pretty terrible to work with since there are no source locations to point to for generated methods, so if there is a missing method for a nested type that the user is using they will get strange errors that happen at unification time instead of when the methods are generated.

Handles type parameters fine & recursive datatypes / datatypes with functions.

Happy to do some additional defensive programming and remove support unless a user annotates a type. It seems like it might be hard foresee all of the various problems that might occur, but it probably isn't as bad as I think. I've already covered only considering value kinded type arguments, and ignoring effects / other kind arguments. 

Finally I think a primitive pointer equivalence check might be good to optimize the equality a tiny bit.

### DIFF
--- a/lib/std/core.kk
+++ b/lib/std/core.kk
@@ -2056,6 +2056,15 @@ pub fun show( s : sslice ) : string
 pub fun show-list( xs : list<a>, show-elem : (a) -> e string ) : e string
   "[" ++ xs.map(show-elem).join(",") ++ "]"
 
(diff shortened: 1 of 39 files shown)


TimWhiting:
Just ran the tests, realized there are quite a few failures (not just changes that need to be updated). Will look at them.

PR #338, opened 2023-05-27
Updated Language Server

TimWhiting:
What the title says, updated to latest dev branch, and got it working, with a couple of improvements.

There is less configuration required now. You can still explicitly specify the Koka executable path, but by default it searches in three locations. First it looks for the Koka repo in the users home directory and looks for the most recent stack build. Then it looks in ~/local/bin, finally it looks for it in the system path. It also selects the current workspace folder as the default folder to be the cwd, and the language server is turned on by default. 

It also uses the virtual file system now, using the contents from the client which includes unsaved changes/edits. This allows for the parse / type errors to show up as you are typing, rather than only when you save.

After this PR or until it gets merged I'll can work on some tests.

Additionally I want to add support for generating standard methods like show / ==. Probably based on my progress in #335 (but unlike that PR not inserted automatically which can be difficult to catch all of the cases with generics & kinds)  

Steps:
- clone the repository 
- install the prerequisites (haskell, stack, etc)
- `stack build`
- start the language extension debug configuration
- Select or create a .kk file and start writing!

Debug configuration is here (add to .vscode/launch.json)
```
{
  // Use IntelliSense to learn about possible attributes.
  // Hover to view descriptions of existing attributes.
  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
  "version": "0.2.0",
  "configurations": [
    {
        "args": [
            "--extensionDevelopmentPath=${workspaceFolder}/support/vscode/koka.language-koka"
        ],
        "name": "Launch Extension",
        "outFiles": [
            "${workspaceFolder}/support/vscode/koka.language-koka/out/**/*.js"
        ],
        "request": "launch",
        "type": "extensionHost"
    }
]
}
```


TimWhiting:
@CaramelFur 
I'm not sure how to get your commits to be fully verified here? I cherry picked them onto the dev branch, and added my commits on top.

TimWhiting:
Update, I've added support to run koka files. You can create a run configuration (just executes koka -e -i{vscodeWorkspaceDirectory}, and pipes the result to the debug console). It also provides codeLenses, so you can run any koka file directly from a hint above the main method.

PR #340, opened 2023-06-21
Explicit struct tag is outside union

TimWhiting:
@daanx 
The union does not contain the tag for the struct. This was causing exceptions because the field access was incorrect.

Was happening with both Optional and Error types. 

Running util/bundle.kk works now. Prior to this change it does not.

Current dev branch emits structs like this

```c
// value type std/core/types/optional
struct kk_std_core_types_Optional {
  kk_box_t value;
};
struct kk_std_core_types_None {
  kk_box_t _unused;
};
struct kk_std_core_types_optional_s {
  kk_value_tag_t _tag;
  union {
    struct kk_std_core_types_Optional Optional;
    struct kk_std_core_types_None None;
    kk_box_t _fields[2];
  } _cons;
};
typedef struct kk_std_core_types_optional_s kk_std_core_types__optional;
static inline kk_std_core_types__optional kk_std_core_types__new_None(kk_context_t* _ctx) {
  kk_std_core_types__optional _con;
  _con._tag = kk_value_tag(2);
  _con._cons._fields[1] = kk_box_null();
  return _con;
}
```

### DIFF
--- a/src/Kind/Infer.hs
+++ b/src/Kind/Infer.hs
@@ -836,7 +836,7 @@ resolveTypeDef isRec recNames (DataType newtp params constructors range vis sort
            conCount       = length conInfos0
            willNeedStructTag   = dataDefIsValue ddef && conCount > 1 && maxMembers >= 1
            extraFields = if (dataDefIsOpen ddef) then 1 {- open datatype tag -}
-                         else if willNeedStructTag then 1 {- explicit struct tag -} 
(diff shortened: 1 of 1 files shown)


TimWhiting:
Fixed by this: https://github.com/koka-lang/koka/commit/ba9433fdb2eb345072e274fb982a697b241668d8

PR #346, opened 2023-07-20
Language server

TimWhiting:
I've reset the commit history here since this PR no longer looks much like the original PR. This PR is updated to the latest Haskell LSP libraries, uses the LSP library's virtual file system to support unsaved changes, and has support for VSCode run configuration of Koka files. Syntax errors are no longer immediately sent to stdout so that it doesn't mess with the server protocol.

I'll go through and make some comments in this PR or in the code for things to note.


TimWhiting:
Several updates fixing a few bugs, reusing modules in memory that have already been typechecked, better completions based on types and better logging and compile & run support via language server.

TimWhiting:
Updated the binary files for the sample release here: https://github.com/TimWhiting/koka/releases/tag/v2.4.2-language-server

TimWhiting:
Updated extension with ability to pass in extra compiler arguments to the run/debug configuration

TimWhiting:
An update:

For small projects, the language server works really well. For large projects or large files, the compiler is really slow, and takes a ton of memory. I've run into some sort of memory leak. Since I'm not super familiar with debugging Haskell memory leaks - especially due to it's laziness, and the memory leak might be in the existing compiler, but just doesn't manifest due to the short runtime of the process, I'm not sure where to go from here. I did look at how to configure the runtime options for Haskell to capture a heap profile, but for whatever reason, when I terminate the process it doesn't write it out - I'm probably sending the wrong signal.

src/Compiler/Compile.hs
```diff
@@ -322,34 +332,32 @@ makeRelativeToPaths paths fname
...
-compileProgram term flags modules compileTarget fname program
-  = runIOErr $ compileProgram' term flags modules compileTarget  fname program
-
+compileProgram :: Terminal -> Flags -> Modules -> Modules -> CompileTarget () -> FilePath -> UserProgram -> IO (Error (Loaded, Maybe FilePath))
+compileProgram term flags modules cachedModules compileTarget fname program
+  = runIOErr $ compileProgram' (const Nothing) term flags modules cachedModules compileTarget fname program
 
-compileProgramFromFile :: Terminal -> Flags -> Modules -> CompileTarget () -> FilePath -> FilePath -> IOErr Loaded
-compileProgramFromFile term flags modules compileTarget rootPath stem
+compileProgramFromFile :: (FilePath -> Maybe (BString, FileTime)) -> Maybe BString -> Terminal -> Flags -> Modules -> Modules -> CompileTarget () -> FilePath -> FilePath -> IOErr (Loaded, Maybe FilePath)
```
TimWhiting, line 353: The main difference from before in compiling a file:

1. There is a parameter `maybeContents` that get's threaded through, which gives in-memory representation of the file (from the editor). This is used instead of the file on disk (it has priority). The virtual file system only keeps files that are open in the editor alive. 
2. There is a parameter `cachedModules` that has modules that have been compiled previously. We can skip inference / typechecking / persisting these. (We do have to keep track of when they do need to actually be persisted - for example when we request the executable)
3. There is a parameter `contents`, that provides the currently selected program's contents according to `maybeContents`, this parameter could be dropped and just use a local variable, I might do that.
4. Compiling returns the filepath of the compiled module in addition to loaded. This allows for us to know the executable path to report to the editor for running the produced executable.

src/Compiler/Compile.hs
```diff
@@ -322,34 +332,32 @@ makeRelativeToPaths paths fname
...
+compileProgramFromFile maybeContents contents term flags modules cachedModules compileTarget rootPath stem
   = do let fname = joinPath rootPath stem
        -- trace ("compileProgramFromFile: " ++ show fname ++ ", modules: " ++ show (map modName modules)) $ return ()
        liftIO $ termPhaseDoc term (color (colorInterpreter (colorScheme flags)) (text "compile:") <+> color (colorSource (colorScheme flags)) (text (normalizeWith '/' fname)))
-       liftIO $ termPhase term ("parsing " ++ fname)
        exist <- liftIO $ doesFileExist fname
        if (exist) then return () else liftError $ errorMsg (errorFileNotFound flags fname)
-       program <- lift $ parseProgramFromFile (semiInsert flags) fname
+       program <- lift $ case contents of { Just x -> return $ parseProgramFromString (semiInsert flags) x fname; _ -> parseProgramFromFile (semiInsert flags) fname}
```
TimWhiting, line 360: Here is where we take priority of the contents of an in memory file.

src/Compiler/Compile.hs
```diff
@@ -435,43 +448,60 @@ compileProgram' term flags modules compileTarget fname program
...
-                                  -> errorMsg (ErrorGeneral (infoRange info) (text "'main' must be declared as a function (fun)"))
-                               [] -> errorMsg (ErrorGeneral rangeNull (text "the type of 'main' must be a function without arguments" <->
-                                                                                      table [(text "expected type", ppType (prettyEnvFromFlags flags) mainType)
-                                                                                            ,(text "inferred type", ppType (prettyEnvFromFlags flags) (head (map infoType infos)))]))
-                               _  -> errorMsg (ErrorGeneral rangeNull (text "found multiple definitions for the 'main' function"))
-             Object -> return (Object,loaded2)
-             Library -> return (Library,loaded2)
-       
-       loaded4 <- liftIO $ codeGen term flags newTarget loaded3
+       (newTarget,loaded3) <- doCodeGen term flags loaded2 compileTarget program coreImports
```
TimWhiting, line 451: I move the code generation piece into a separate function below in order to allow us to perform code generation lazily when needed for dependencies when changing from InMemory to Object.

src/Compiler/Compile.hs
```diff
@@ -435,43 +448,60 @@ compileProgram' term flags modules compileTarget fname program
...
+       (loaded4, outFile) <- liftIO $ case newTarget of
+            InMemory -> return (loaded3{loadedModule = (loadedModule loaded3){modOutputTime = Nothing}}, Nothing)
```
TimWhiting, line 453: I moved the evaluation part of `codeGen` up here, as it seems strange for `codeGen` to be in charge of evaluation. I'm not sure if it is best here (it could be even higher). Since we report the executable location, this is now possible to do here.

src/Compiler/Compile.hs
```diff
@@ -584,9 +614,14 @@ searchModule flags currentDir name
                          Nothing -> searchPackageIface flags currentDir Nothing name
                          Just iface -> return (Just iface)
 
+getCurrentFileTime :: FilePath ->  (FilePath -> Maybe (BString, FileTime)) -> IO FileTime
+getCurrentFileTime fp maybeContents =
+  case maybeContents fp of
+    Just (_, t) -> return t
+    Nothing -> getFileTimeOrCurrent fp
```
TimWhiting, line 621: This method gets the current file time, giving preference to a `FileTime` reported by the virtual file system.

src/Compiler/Compile.hs
```diff
@@ -642,31 +677,56 @@ resolveModule term flags currentDir modules mimp
 
     where
       name = impFullName mimp
+      
+      tryLoadFromCache :: FilePath -> FilePath -> FilePath -> IOErr (Maybe (Module, Modules))
+      tryLoadFromCache iface root stem
```
TimWhiting, line 682: This method, loads from the `cachedModules`, subsequently calling `loadFromModule` to load all depedencies.

TimWhiting: loadFromIface and loadFromSource are adjusted accordingly, to defer to this method when the module is not already in loaded, and prior to actually loading from iface / source

src/Compiler/Compile.hs
```diff
@@ -696,15 +763,30 @@ resolveModule term flags currentDir modules mimp
...
-               then loadFromSource resolved1 root source -- load from source after all
-               else do liftIO $ copyPkgIFaceToOutputDir term flags iface (modCore mod) (modPackageQPath mod) imports
-                       let allmods = addOrReplaceModule mod resolved1
-                       return (mod{ modSourcePath = joinPath root source }, allmods)
+               then loadFromSource True resolved1 root source -- load from source after all
+               else 
+                let allmods = addOrReplaceModule mod resolved1
+                    result = (mod{ modSourcePath = joinPath root source }, allmods)
+                in case compileTarget of
+                  InMemory -> return result
```
TimWhiting, line 776: Here is the main difference in loadFromModule. We do not actually copy ifaces to output directory and we do not do code generation unless we need to (based on the last time the module was inferred and the last time it was written).

src/Compiler/Compile.hs
```diff
@@ -813,13 +895,13 @@ inferCheck :: Loaded -> Flags -> Int -> [Core.Import] -> UserProgram -> Error (L
...
     do -- kind inference
-       
+
        (defs, kgamma, synonyms, newtypes, constructors, coreProgram, mbRangeMap0)
          <- inferKinds
               (isValueFromFlags flags)
               (colorSchemeFromFlags flags)
               (platform flags)
-              (if (outHtml flags > 0) then Just rangeMapNew else Nothing)
+              (if (outHtml flags > 0 || genRangeMap flags) then Just rangeMapNew else Nothing)
```
TimWhiting, line 904: We need to generate range maps for the language server.

src/Compiler/Options.hs
```diff
@@ -632,6 +644,18 @@ getOptions extra
        args <- getArgs
        processOptions flagsNull (env ++ words extra ++ args)
 
+updateFlagsFromArgs :: Flags -> String -> Maybe Flags
+updateFlagsFromArgs flags0 args =
```
TimWhiting, line 648: This is used for parsing additional arguments from the language server's run command.

src/Core/Parse.hs
```diff
@@ -46,7 +46,7 @@ type ParseInlines = Gamma -> Error [InlineDef]
 parseCore :: FilePath -> IO (Error (Core, ParseInlines))
 parseCore fname
   = do input <- readInput fname
-       return (lexParse False (requalify . allowDotIds) program fname 1 input)
+       return $ ignoreSyntaxWarnings $ lexParse False (requalify . allowDotIds) program fname 1 input
```
TimWhiting, line 49: lexParse is changed to add syntax warnings to it's state, and report them in batch after parsing. We assume core ifaces don't have syntax warnings.

src/LanguageServer/Handler/Commands.hs
```diff
@@ -0,0 +1,63 @@
...
+targetFlag t f
+    = case lookup t targets of
+        Just update -> update f
+        Nothing     -> f
+
+commandHandler :: Handlers LSM
+commandHandler = requestHandler J.SMethod_WorkspaceExecuteCommand $ \req resp -> do
+  flags <- getFlags
+  let J.ExecuteCommandParams _ command commandParams = req ^. J.params
+  if command == "koka/genCode" then
```
TimWhiting, line 38: This is the genCode command, which instructs the language server to actually output an executable and report the executable's location

src/LanguageServer/Handler/Commands.hs
```diff
@@ -0,0 +1,63 @@
...
+    case commandParams of
+      Just [Json.String filePath, Json.String additionalArgs] -> do
+        term <- getTerminal
+        newFlags <-  case updateFlagsFromArgs flags (T.unpack additionalArgs) of
```
TimWhiting, line 42: Parse additional command line arguments

src/LanguageServer/Handler/Commands.hs
```diff
@@ -0,0 +1,63 @@
...
+          Just flags' -> return flags'
+          Nothing -> do
+            doc <- liftIO (commandLineHelp flags)
+            sendNotification J.SMethod_WindowLogMessage $ J.LogMessageParams J.MessageType_Error $ T.pack "Invalid arguments " <> additionalArgs
+            liftIO $ termPhaseDoc term doc
+            return flags
+        withIndefiniteProgress (T.pack "Compiling " <> filePath) J.NotCancellable $ do
+          res <- recompileFile (Executable (newName "main") ()) (J.filePathToUri $ T.unpack filePath) Nothing True newFlags
```
TimWhiting, line 50: Recompile with Executable compileTarget

src/Lib/Printer.hs
```diff
@@ -214,6 +216,47 @@ instance Printer AnsiPrinter where
   setUnderline p u      = unit $ ansiSetConsole p (\con -> con{ underline = u })
 
 
+data AnsiStringPrinter = AnsiString (Var AnsiConsole) (Var String)
```
TimWhiting, line 219: Added a printer that prints color, but to a string instead of IO, so that we can show colored output in the language server.

src/Syntax/Parse.hs
```diff
@@ -60,12 +61,13 @@ import Syntax.Lexeme
...
 import Syntax.Layout  ( layout )
 import Syntax.Promote ( promote, promoteType, quantify, promoteFree )
+import Common.ColorScheme (defaultColorScheme)
 
 -----------------------------------------------------------
 -- Parser on token stream
 -----------------------------------------------------------
 
-type LexParser a  = Parsec [Lexeme] () a -- GenParser Lexeme () a
+type LexParser a  = Parsec [Lexeme] [(String, Range)] a -- GenParser Lexeme () a
```
TimWhiting, line 70: The main change here is to report syntax warnings in batch, and not cause termination. This way we can report syntax errors in the language server.

src/Type/Infer.hs
```diff
@@ -1288,7 +1289,7 @@ inferApp propagated expect fun nargs rng
                                     cargs = [Core.Var (Core.TName var (Core.typeOf arg)) Core.InfoNone | (var,(_,arg)) <- vargs]
                                 if (Core.isTotal fcore)
                                 then return (Core.makeLet defs (coreApp fcore cargs))
-                                else do fname <- uniqueName "fun"
+                                else do fname <- uniqueName "fct"
```
TimWhiting, line 1292: Core was having problems parsing names with fun occasionally.

src/Type/Pretty.hs
```diff
@@ -192,8 +192,9 @@ ppSchemeEffect env tp
 
 prettyDefFunType :: Env -> [ParamInfo] -> Scheme -> Doc
 prettyDefFunType env pinfos tp 
-  = let (Just params,pre,post) = ppDeclType env pinfos tp
-    in pre <.> parens (commaSep (map ppParam params)) <+> text "->" <+> post
+  = case ppDeclType env pinfos tp of
+      (Just params,pre,post) -> pre <.> parens (commaSep (map ppParam params)) <+> text "->" <+> post
+      (Nothing,pre,post) -> pre <+> text "()" <+> text "->" <+> post
```
TimWhiting, line 197: The nothing case does happen occasionally, when showing function types on hover.

TimWhiting (commented):
Hi @daanx 

I've added a few comments to this PR, and did some cleanup. This PR should be a lot cleaner now. I believe the current approach in `Compile.hs` should make a lot more sense than my prior approach since I understand it better now, but I haven't used it as extensively yet. I believe the memory issues to be less of a problem than mentioned previously, though likely still present.

TimWhiting:
Latest prebundled release assets that I have generated:
Several improvements since the last release.

https://github.com/TimWhiting/koka/releases/tag/v2.4.3-lang-server

TimWhiting:
With the most recent changes, there are a lot of improvements, including some fixes for windows paths.

Precompiled binaries and the vscode extension are currently available here: https://github.com/TimWhiting/koka/releases/tag/v2.4.3-lang-server7

I'm working with Daan to get the changes integrated in the next release, in addition to a few other fixes in recent PRs (which are included in the precompiled release linked above).

TimWhiting:
Moving to lang-server2.1 branch

PR #364, opened 2023-09-22
Docs update

TimWhiting:
Add fip fbip and tail keywords to documentation, spec and bison/yacc, as well as some documentation on named handlers.

### DIFF
--- a/doc/spec/spec.kk.md
+++ b/doc/spec/spec.kk.md
@@ -108,11 +108,13 @@ grammar will draw it's lexemes from the _lex_ production.
 |              | &bar; | `match` &bar; `return` &bar; `with` &bar; `in`                                       |                         |
 |              | &bar; | `handle` &bar; `handler` &bar; `mask`                                                   |                         |
 |              | &bar; | `ctl` &bar; `final` &bar; `raw`                                     |                         |
-|              | &bar; | `override` &bar; `named`                                               |                         |
+|              | &bar; | `override` &bar; `named`                                        |                         |
 |              | &bar; | `interface` &bar; `break` &bar; `continue` &bar; `unsafe`        | (future reserved words) |
 | &nbsp;       |       |                                                                           |                         |
 | _specialid_  | ::=   | `co` &bar; `rec` &bar; `open` &bar; `extend` &bar; `behind`   |                         |
-|              | &bar; | `linear` &bar; `value` &bar; `reference`                            |                         |
+|              | &bar; | `linear` &bar; `scoped` &bar;                                 |                     |
+|              | &bar  | `value` &bar; `reference` &bar; `ref`           |                         |
+|              | &bar; | `fip` &bar; `fbip` &bar; `tail`                                                   |                         |
 |              | &bar; | `inline` &bar; `noinline` &bar; `initially` &bar; `finally`      |                         |
 |              | &bar; | `js` &bar; `c` &bar; `cs` &bar; `file`                           |                         |
 {.grammar .lex}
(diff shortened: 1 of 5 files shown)


doc/spec/spec.kk.md
```diff
@@ -113,6 +113,7 @@ grammar will draw it's lexemes from the _lex_ production.
 | &nbsp;       |       |                                                                           |                         |
```
chtenb, line 113: Is scoped missing here?

chtenb (commented):
This would be great to have merged. I was reviewing a treesitter implementation for Koka (over at https://github.com/helix-editor/helix/pull/8727), when I noticed that the grammar docs did not align with all of the present features.

TimWhiting:
Let me know if you see any more discrepancies

support/pygments/koka.py
```diff
@@ -25,19 +25,19 @@ class KokaLexer(RegexLexer):
...
         'if', 'then', 'else', 'elif', 
         'match', 'return', 'with', 'in',
         'handle', 'handler', 'mask',
         'ctl', 'final', 'raw',
-        'override', 'named',
+        'override', 'named', 'scoped',
         'interface', 'break', 'continue', 'unsafe',
         'co', 'rec', 'open', 'extend', 'behind',
-        'linear', 'value', 'reference', 
+        'linear', 'value', 'reference',
```
chtenb, line 40: Not entirely sure about this, but isn't `ref` (a few lines down from here) a keyword too rather than a `builtin`?

support/pygments/koka.py
```diff
@@ -25,19 +25,19 @@ class KokaLexer(RegexLexer):
 
     keywords = [
         'infix', 'infixr', 'infixl',
-        'module', 'import', 'as',
-        'pub', 'abstract',
+        'module', 'import', 'as', 'mod'
+        'pub', 'abstract', 'ctx'
```
chtenb, line 29: What does `ctx` mean as a keyword?

chtenb (commented):
Noticed two other things

support/pygments/koka.py
TimWhiting: Actually it is a built-in identifier. It is also the way to create a mutable reference cell.

support/pygments/koka.py
TimWhiting: https://www.microsoft.com/en-us/research/uploads/prod/2023/07/fiptree-tr-v3.pdf

First class constructor contexts. You can start at section 3 of the paper, if you just want to get the high level idea of it. If you want to read the rest of the paper, and not just section 3 I would start at the beginning.

chtenb: Interesting. I see it is used only in the dev branch, so I hadn't spotted it before. Shouldn't this keyword also be mentioned in the parser and lexer definitions?

support/pygments/koka.py
chtenb: Oh right, so it has a different meaning depending on the context. I guess that is not easy to encode in the pygments file, so the identifier was chosen here.

TimWhiting: Yeah, it's a tricky one.

support/pygments/koka.py
TimWhiting: Yeah, I'll have to look at the parser to see how it is done. It's used both in a type context, and as a keyword in the creation of a constructor context.

chtenb:
Over at https://github.com/helix-editor/helix/pull/8727/files @mtoohey31 noticed that the `mod` keyword isn't found on the dev or master branches. Where does this keyword come from?

TimWhiting:
I should have checked. For whatever reason I thought Koka had shortened module to mod with all of the other keywords that got shortened. My bad.

PR #367, opened 2023-10-31
Add comments and smart indentation for (neo)vim

oxalica:
It's mostly modified from [runtime/indent/lua.vim](https://github.com/vim/vim/blob/1bf1bf569b96d2f9b28e0cce0968ffbf2fb80aac/runtime/indent/lua.vim)

Regex is not strong enough to parse all cases, and I try to guess the indentation for common cases. Currently it auto-indents for all `fun/fn` following a `:`, assuming that most functions with an explicit return type are NOT defined inline.

I also renamed `ftdetect/kokafiletype.vim` to `ftdetect/koka.vim` to fit the naming convention.


daanx:
Thanks so much!

PR #372, opened 2023-11-07
fix extend type

TimWhiting:
Previously, accessors were being generated when adding a single constructor to an open type (since all of the added constructors had the same fields), creating a dummy constructor with no fields is a workaround, but the real fix is to not generate accessors for open or extend types.


daanx:
Thanks Tim !

PR #382, opened 2023-12-05
fix hidden operator name

TimWhiting:
This is an issue I ran into when an operator was specialized to a tail call.
There was a fix that was commented out (wasn't working fully), but I fixed the fix. :) - The dash also needed to be ascii encoded.

### DIFF
--- a/src/Common/Name.hs
+++ b/src/Common/Name.hs
@@ -289,14 +289,13 @@ isHiddenName name
       _       -> False
 
 makeHiddenName s name
-  = {-case nameId xname of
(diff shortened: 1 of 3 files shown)


TimWhiting:
@daanx I added a simple reproducible test for this one. Not sure what to do about the output of the test though. The problem is when loading an interface file, which doesn't happen for a regular test compile. (You should be able to reproduce by removing the changes in Name.hs, and then compiling test/syntax/trmc-operator.kk and then test/syntax/trmc-operator2.kk.

daanx:
Thanks! However, I am currently redoing names for the `dev-overload` branch with the locally qualified names and I will try to find a more robust solution -- so for now, I think it is better to not merge yet.

PR #384, opened 2023-12-06
Libuv

TimWhiting:
Implemented: (Low Level, Callback, LibUV wrapper)
Streams
File I/O
Net / TCP I/O
Timer
FS Poll
FS Event
Signals
TTY

Implemented: Async API
Basic File I/O
Basic TCP I/O
Timer (Also implemented for js and wasm)

Implemented: llhttp parser

Missing LibUV: UDP, Threading, Process, uv async channels, Prepare/Check/Idle

Missing (external libraries): SSL!


lib/std/core.kk
```diff
@@ -72,6 +72,12 @@ pub alias io-total = <ndet,console,net,fsys,ui,st<global>>
 // The `:io-noexn` effect is used for functions that perform arbitrary I/O operations, but raise no exceptions
 pub alias io-noexn = <div,io-total>
 
+// The `:event-loop` effect signifies that a function requires an initialized event loop
+pub type event-loop :: X
```
TimWhiting: Here is one major change. Adding event loop to the core effects. 
The async library requires that there be an event loop around it, or at the very least a timer that can call callbacks.

In principle we could just use an implicit parameter for the handler for the async effect, to get the event loop and wrap it.

The problem with that approach is that if you use any libuv apis directly in callback style without the async library then you have to manually remember to initialize the event loop. 
We could in principle make a custom error and detect uninitialized event loops at runtime, but that would introduce runtime overhead for each libuv call that uses callbacks, so I opted to create this new effect instead.

I will note that I did use the implicit parameter approach for the timer requirements of the async lib so that in principle you can swap out the libuv library for some other event loop library.

cognivore:
Sorry for a silly question — would you say trunk of libuv integraiton is in `dev-uv` branch now?

TimWhiting:
Yes. This is just a reference point now and should be considered abandoned.

daanx:
Yes, currently the `dev-uv` branch is for the new async api and we are cherry picking existing code from other PR's / branches when it comes up.  I plan to merge `dev-uv` back into `dev` and create a new release once the basic things are working well (timer, file, network) and extend from there until we can write http(s) servers :-)  I think once the basics are set up well it will be easier to contribute and extend for others.

PR #403, opened 2023-12-28
Fix vector out of bounds for JS backend

mtoohey31:
Hello! I noticed that vector out of bounds checks don't appear to be working on either the C or JS backends. This pull request fixes them on the JS backend.

I haven't figured out how to fix them on the C backend yet. I was looking for an example of throwing a Koka error from C, but haven't found one. Is this possible? If it is and you can point me in the right direction in terms of how to do so, I'd be happy to fix the C backend too.


mtoohey31:
Closing since #410 takes care of this.

PR #410, opened 2023-12-30
Vector OOB fix

TimWhiting:
#403, but with additional C support, and a test


daanx:
Thanks Tim! I fixed it by not using externals (in the future we should look at better codegen overall for vectors and bytes). 
Added your test as well.

Issue #416, opened 2023-12-31
vector-init does not work correctly for control effects


anfelor:
In #388, the `vector-init` function was changed to allow an arbitrary effect on the initialization function. While this is a welcome change for most effects, it does not play well with control effects. Consider the following program:

```
effect choice
  ctl flip() : bool

fun test-vector()
   with handler
     ctl flip()
       val v1 = resume(True)
       val v2 = resume(False)
       return v1.append(v2)
   with return(v) [v]
   vector-init(5, fn(i) -> flip())

fun test-list() // Just like test-vector but for lists
   with handler
     ctl flip()
       val v1 = resume(True)
       val v2 = resume(False)
       return v1.append(v2)
   with return(v) [v]
   list(0, 4, fn(i) -> flip())

fun main()
   val vs = test-vector()
   // val vs = test-list()
   println(vs.length())
   vs.foreach(fn(v) print(v.length()))
   return ()
```

The expected behaviour is that `test-vector` should return `2^5` vectors of length `5` (containing all possible combinations of `True` and `False` elements in a 5-element vector) -- just like the `test-list` function returns the corresponding lists. But in Koka 2.6.0, `test-vector` only returns two zero-element vectors.

TimWhiting:
Ah, we would need an if yielding check in the hand written c code for this.

TimWhiting:
I'm wondering whether this would be an appropriate solution @daanx:

```koka
// Create a new vector of length `n`  with initial elements given by function `f` .
pub fun vector-init( ^n : int, f : int -> e a ) : e vector<a>
  val v = unsafe-vector(n)
  forz( 0.ssize_t, n.ssize_t ) fn(i)
    unsafe-assign(v,i,f(i.int))
  v
```

daanx:
Thanks so much! Fixed now.

PR #420, opened 2024-01-04
fix vector init

TimWhiting:
Fixes: #416 
Vector init wasn't working properly for control effects. I don't know if you'd rather just make the effect `io-noexn`, but it seems to me that we would at least want to support `st<>` or `ref<>` as well. Either way it probably should be documented that using multiple resumptions with this API is bad practice because it doesn't copy the vector.

I went through the rest of core.kk and made sure there were no more functions being passed to c that have polymorphic effect types.


daanx:
Thanks for the help Tim! I ended up with a different approach but also used the new `vector-init` in koka itself to deal with control effects.

PR #426, opened 2024-01-06
Gen show / (==) using implicits

TimWhiting:


TimWhiting:
It looks like there is now an issue with generating these due to the expectation in core that (==) is total. It should be, but in general you cannot guarantee non-divergence.

TimWhiting:
This also adds the start of a pretty printing module for user syntax which I found useful while debugging this.

TimWhiting:
Superceded by #622

PR #445, opened 2024-01-24
Fix error in docs

chtenb:
The docs mention that `p( age = age + 1)` is valid syntax for the copy constructor, which I found slightly surprising, since in other places `age` is properly called with an argument, like `p.age`. As it doesn't actually seem to work either, I assume it's a mistake/typo in the docs.


TimWhiting:
Thanks!

Issue #457, opened 2024-02-01
std/text/unicode width does not return column-width 2 for emojis


erf:
I did expect the `width` function to return 2 for emojis when using the [EastAsianWidth.txt](https://www.unicode.org/Public/15.0.0/ucd/EastAsianWidth.txt) file.

```
  println(width("👾"))
```

this returns 1

Is this method supposed to work similar to the [display_width](
https://codeberg.org/dude_the_builder/ziglyph/src/commit/ac50ab06c91d2dd632ff4573c035dafe3b374aba/src/display_width.zig#L31) method of `ziglyph` or similar to this [Python wcwidth spesification](https://wcwidth.readthedocs.io/en/latest/specs.html) ? That is to give the rendered column width for modern terminal emulators using the latest Unicode standard?

TimWhiting:
This is what the standard library documentation says:
```koka
// Return the column-width of a unicode character.
// Equivalent to ``wcwidth``
pub fun char/width( c : char ) : int {
  if (zero-widths.force.contains(c.int)) then 0
  elif (asian-wide.force.contains(c.int)) then 2
  else 1
}

// Return the total column-width of a string.
pub fun string/width( s : string ) : int {
  var total := 0
  s.foreach( fn(c) {
    total := total + c.width
  })
  total
}
```
So yes, I believe the intent is for terminal emulators as in the python wcwidth spec, however I'm not certain if it is currently up to date.

Also I would expect the following to print two utf16 characters, but it only does one utf32 character. I guess I'm less certain on the intended underlying representation for characters. I'll have to ask Daan.
 ``` "👾".slice.foreach(fn(c) c.println)```

erf:
I'll just link this article here. It's a good read with some valuable links

https://mitchellh.com/writing/grapheme-clusters-in-terminals

TimWhiting:
Thanks for the link!

TimWhiting:
Great post. I'll have to look at the algorithm he references.

erf:
Yeah i'm a beta tester on the Ghostty terminal (it's great!), and they have implemented [Mode 2027](https://github.com/contour-terminal/terminal-unicode-core) for proper Unicode handling

PR #461, opened 2024-02-03
Update unicode

### LINKED ISSUE #458: std/text/unicode graphemes does not return a list of grapheme clusters to iterate
I thought `graphemes("hi❤️‍🔥")`

would return the list: `["h", "i", "❤️‍🔥"]`, a list of grapheme clusters that i could iterate with:

```
  l.foreach fn(c)
    println(c)
```

which would print out single grapheme clusters like: 
```
h
i
❤️‍🔥
```

also if i print `l.length` now it returns `6`, i wish there was a function which would return the number of grapheme clusters like 3 in this case.

I'm new to `koka` and these libraries so sorry if i've mistaken the usage.

TimWhiting:
Fixes: #458 & #457 

There is more work to do to grapheme clusters completely correct, but this is a step in that direction.

See https://www.unicode.org/reports/tr29/#Grapheme_Cluster_Boundary_Rules for what needs to be done to support this properly.


erf:
BTW i think the latest Unicode version in use is 15.1 (16 is in draft)

https://www.unicode.org/Public/UNIDATA/EastAsianWidth.txt

TimWhiting:
> i think the latest Unicode version in use is 15.1 (16 is in draft)
> 
> https://www.unicode.org/Public/UNIDATA/EastAsianWidth.txt

Thanks for the pointer to a link that doesn't have a hard-coded version in it. I've created a python script so we can make updating this part of our release process.

erf:
I'm making a script like that myself for my Dart vi-like editor [vid](https://github.com/erf/vid) . BTW beware the format has changed slightly with the "W" in a different place

TimWhiting:
I came from the Dart community myself 😄. Glad to see that other people have found Koka.

TimWhiting:
> It seems like a great language I'd like to explore ! Although it might still be in a early stage in regards of libraries etc.

Yes it is. Libraries can't be done alone though; they require a community. We are getting more active at accepting contributions. I hope you can see the potential for Koka to grow, and become part of the community - at least as an enthusiast, even if it isn't your day job yet :).

TimWhiting:
One difficult thing about Koka and creating the standard libraries is that Koka includes a bunch of non-standard features that directly affect api design. We are still learning what idiomatic Koka code looks like. You can help with that :).

erf:
I definitely can see the potential, although I've only recently started looking into it. I do have a lot of things going on right now, so not sure how much i can contribute at the moment, but will keep an eye on it. In my experience building a community requires a strong leader - especially in the initial phase. Keep up the good work :)

TimWhiting:
I've updated this PR with common families of characters from [DerivedCoreProperties](https://www.unicode.org/Public/UNIDATA/DerivedCoreProperties.txt). 

Unfortunately, it seems to cause the compiler to hang. I've noticed that very large lists can cause trouble for the compiler.

PR #479, opened 2024-03-20
Install pacman dependencies only if needed

egmaleta:
This will skip the download and reinstallation of already installed deps


TimWhiting:
Thanks!

PR #490, opened 2024-04-10
The start of a testing library idea

TimWhiting:


TimWhiting:
I'm going to close this. I think this could use some more community feedback. I've started a community organization on github at (https://github.com/koka-community). There, we can iterate and experiment more without worrying about backwards compatibility.

PR #491, opened 2024-04-10
Allow named arguments anywhere

TimWhiting:
The main motivation for this feature is to allow trailing lambdas to come after named arguments. I've run into this a lot.

```koka
f(1, x="something") fn()
   "something else"
```
(Currently an error).

Essentially, instead of thinking of formal parameters as named or fixed, we consider them required or optional, with implicit parameters being required parameters, but optional as arguments.
Arguments are still thought of as named or fixed.

Subsumption proceeds as follows:

After matching the named arguments to formal parameters, the fixed arguments must match the ordering of the remaining parameters, with optional parameters staying optional.

In making these changes I noticed that the `syntax/run` test subdirectory was not actually being run, since it inherited a `-l` flag from the `syntax` directory. I moved those tests into a `syntax/no-run` subdirectory to prevent the conflicting flags.

Alternatively we could adjust the trailing lambda desugaring to put it prior to all named arguments. See PR #533 

The main advantage of this particular solution is that it also allows for function "configuration" to happen prior to giving common arguments that would cause more indentation and visual clutter / noise if named.
```
div(
  classes="x,y,z", 
  div(
...
))
```

Could be written like this, but becomes noisy and causes deeper indentation (depends how you format it)
```
div(
  classes="x,y,z", 
  child=div(
...
))
```


test/syntax/run/tail.kk.out
TimWhiting, line 1: There seems to be a bug (that is currently also on dev, where implicit resolution considers but doesn't accept functions which would match using a default for an optional parameters).

TimWhiting: Fixed in #493

src/Type/Infer.hs
```diff
@@ -2034,16 +2034,12 @@ isAnnot _                     = False
 
 splitNamedArgs :: Ranged e => [(Maybe (Name,Range),e)] -> Inf ([e],[((Name,Range),e)])
 splitNamedArgs nargs
-  = do let (nfixed,rest)      = span (isNothing . fst) nargs
-           (nnamed,morefixed) = span (not . isNothing . fst) rest
+  = do let nfixed      = filter (isNothing . fst) nargs
+           nnamed      = filter (not . isNothing . fst) nargs
            fixed              = map snd nfixed
```
TimWhiting, line 2042: Instead of using span, we just filter by whether they have a name.

src/Type/Infer.hs
```diff
@@ -2034,16 +2034,12 @@ isAnnot _                     = False
...
            named              = [((name,rng),expr) | (Just (name,rng),expr) <- nnamed]
        -- check that named arguments all come after the positional ones
-       case morefixed of
-         [] -> return ()
-         (arg:_) -> infError (getRange (snd arg)) (text "positional arguments cannot follow named arguments")
```
TimWhiting, line 2047: No error is needed any more!

src/Type/Unify.hs
```diff
@@ -138,10 +138,11 @@ matchArguments matchSome range free tp fixed named mbExpResTp
               then unifyError NoMatch
               else do -- trace (" matchArguments: " ++ show (map pretty pars, map pretty fixed, map pretty named)) $ return ()
                       -- subsume fixed parameters
-                      let (fpars,npars) = splitAt (length fixed) pars
+                      let parsNotNamedArg = filter (\(nm,tp) -> nm `notElem` map fst named) pars
+                      let (fpars,rest) = splitAt (length fixed) parsNotNamedArg
```
TimWhiting, line 142: The only other change really is during unification removing all the supplied named arguments from parameters before determining which ones are fixed.

TimWhiting: Other than two more one line changes just below this, all of the rest of the changes are in updating tests.

daanx:
Thanks Tim -- very cool PR!

PR #493, opened 2024-04-11
Eta expand functions with optional parameters when resolving implicits

TimWhiting:
When using a function with an optional parameter as implicit, the types do not unify. Eta-expanding the function fixes the issue. This is particularly important due to `double/show` having the default parameter for precision.

### DIFF
--- a/lib/std/core/console.kk
+++ b/lib/std/core/console.kk
@@ -71,7 +71,7 @@ pub fun string/print(s : string)
   prints(s)
 
 // Print a value that has a `show` function
-pub fun show/print( x : a, ?show : a -> string ) : console ()
+pub fun show/print( x : a, ?show : a -> <console|e> string ) : <console|e> ()
   prints(x.show)
 
 // Print a string to the console, including a final newline character.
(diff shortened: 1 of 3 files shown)


daanx:
Hi Tim -- thanks for the PR, I think it looks good. However, I have generalized this idea a bit and added code to automatically eta-expand any _variable_ argument that has optional parameters but the expected type does not. For example, in:
```koka
fun plus(x:int,y:int,z:int=1)
  x+y+z

pub fun main()
  [1,2,3].foldl(0,plus)
```
I think the commit 123d3e7 also handles what this PR is doing -- but I'm not 100% sure; maybe you can see if it covers your use cases?

TimWhiting:
@daanx 
This still fails: `stack run koka -- -e test/syntax/run/tail`. Also apparently the additional config.json options in the run directory are not replacing or being added the options in the parent directory, so the `run` tests aren't actually being run.

daanx:
Ah,  I think this is a different problem -- namely, the `main` that is generated does a `println` on a `float64` but it doesn't import `std/num/float64`. This is something we should fix as well.

daanx:
I think the import problem is fixed now in the latest `dev` (commit f05b93c14dcdfbefc18eb3484d31cb4b087e5c44) . (the `tail` tests work)

PR #494, opened 2024-04-11
Support external managed memory

TimWhiting:
I've been thinking a bit about automated generation of C wrappers (along the lines of a port of Dart's ffigen), but of course first we need a way of managing external ffi resources in Koka. After searching in `kklib` for a bit, I found the `kk_cptr*` functions which don't currently have a corresponding Koka api. Here is an initial Koka api for these functions, allowing us to manage allocated C memory completely from Koka.

### DIFF
--- a/lib/std/core/cextern.kk
+++ b/lib/std/core/cextern.kk
@@ -7,6 +7,78 @@ import std/core/int
 extern import
   c header-file "inline/cextern.h"
 
+// TIM: Some thoughts / notes on wrapping C libraries and C / extern memory management
+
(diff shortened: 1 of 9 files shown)


TimWhiting:
Marking this as draft: I've since made some updates here, which I need to migrate back to this PR: https://github.com/koka-community/std/blob/main/std/cextern.kk

TimWhiting:
By the way, I have a version of the libuv branch, which uses this api, and automatically generated ffi bindings (adapted from dart's `ffigen`). This allows us to write almost all of the external binding code in koka.

PR #495, opened 2024-04-12
Add implicits hint

TimWhiting:
Report missing implicit definitions when an implicit cannot be found, needed in all candidates.
Also hint to reorder implicit definitions prior to functions that need them.


daanx:
Thanks Tim!

PR #513, opened 2024-05-03
Fix duplicate label issue, include index for duplicate effect labels

### LINKED ISSUE #511: Mask is ignored when adding `exn` effect
Consider the following example, where we use two different handlers to catch two `abort` effects.

```koka
effect abort
  ctl abort() : a

fun catch'(f : () -> <abort|e> a, h : () -> e a) : e a
  with ctl abort() { h() }
  f()

fun maybe(f : () -> <abort|e> a) : e maybe<a>
  with ctl abort() { Nothing }
  Just(f())

type abc { A; B; C }

fun show(abc : abc)
  match abc
    A -> "A"
    B -> "B"
    C -> "C"

fun abortA(abc : abc)
  match abc
    A -> abort()
    _ -> abc

fun abortB(abc : abc)
  match abc
    B -> abort()
    _ -> abc

fun distinct(abc : abc) : <abort,abort> abc
  val n = mask<abort>
    abortA(abc)
  abortB(n)

fun main() : <console> ()
  println(maybe(fn() catch'(fn() distinct(A), fn() C)))
  println(maybe(fn() catch'(fn() distinct(B), fn() C)))
  println(maybe(fn() catch'(fn() distinct(C), fn() C)))
```

This program runs as expected, where `abortA()` is caught by the `maybe` handler and `abortB()` is caught by the `catch'` handler. The output thus is:

```
Nothing
Just(C)
Just(C)
```

However, if we merely change the type signature to `fun main() : <console, exn> ()`, we obtain:

```
Just(C)
Just(C)
Just(C)
```

where now `abortA()` is also caught by the `catch'` handler (and the mask is seemingly ignored).

TimWhiting:
Fixes #511 

@daanx, I understand if you would like to redo this yourself to make it cleaner, but this should illustrate the issue / solution.

Specifically `@evv-index` does not include an index for duplicate labels. This is fine for most effect expressions, since the effect row is incrementally adjusted. However, for the general `open` case where you use a vector of indices and there are multiple labels of the same kind, we get invalid offsets. This would happen with or without the code in `Simplify.hs` which turns them into constant offsets. 

In general I think the types for effect rows need to represent duplicate label indices more explicitly (when extracting the ordered effect), so I defaulted to including the count of duplicate labels (pseudo index) in `extractOrderedEffect`. However, the indices aren't really important till after type checking, so there are a lot of places where I needed to discard the indices. I don't know if you would rather do something to represent this invariant more explicitly in the type, or if you would rather create another function for when you don't care about indices, since there are only a few places that do.

In particular the issue is illustrated by the following short code snippet from the core output. (from the example in #511)
```koka
(std/core/hnd/@open1(
                        (std/core/vector/unvlist(
                       // Here is the problem both of these indices are the same!!!!
                        (std/core/types/Cons((std/core/types/@make-ssize_t(0)), 
                        (std/core/types/Cons((std/core/types/@make-ssize_t(0)), 
                        std/core/types/Nil)))))
```

### DIFF
--- a/lib/std/core/inline/hnd.c
+++ b/lib/std/core/inline/hnd.c
@@ -73,14 +73,18 @@ static kk_std_core_hnd__ev* kk_evv_as_vec(kk_evv_t evv, kk_ssize_t* len, kk_std_
 // }
 
 
-kk_ssize_t kk_evv_index( struct kk_std_core_hnd_Htag htag, kk_context_t* ctx ) {
+kk_ssize_t kk_evv_index(int32_t index, struct kk_std_core_hnd_Htag htag, kk_context_t* ctx ) {
   // todo: drop htag?
   kk_ssize_t len;
+  kk_ssize_t off = index;
   kk_std_core_hnd__ev single;
   kk_std_core_hnd__ev* vec = kk_evv_as_vec(ctx->evv,&len,&single,ctx);
   for(kk_ssize_t i = 0; i < len; i++) {
     struct kk_std_core_hnd_Ev* ev = kk_std_core_hnd__as_Ev(vec[i],ctx);
-    if (kk_string_cmp_borrow(htag.tagname,ev->htag.tagname,ctx) <= 0) return i; // break on insertion point
+    if (kk_string_cmp_borrow(htag.tagname,ev->htag.tagname,ctx) <= 0) {
+      if (off == 0) return i; // break on insertion point
+      off--; // adjust for duplicate label
+    }
   }
   //string_t evvs = kk_evv_show(dup_datatype_as(kk_evv_t,ctx->evv),ctx);
   //fatal_error(EFAULT,"cannot find tag '%s' in: %s", string_cbuf_borrow(htag.htag), string_cbuf_borrow(evvs));
--- a/src/Core/OpenResolve.hs
+++ b/src/Core/OpenResolve.hs
@@ -133,7 +133,7 @@ resOpen (Env penv gamma) eopen effFrom effTo tpFrom tpTo@(TFun targs _ tres) exp
               then -- all handled effect match, just use expr
                    trace "masking? " $ expr
               else -} failure $ ("Core.openResolve.resOpen: todo: masking handled effect: " ++ show (ppType penv effFrom))
-       else if (matchType tlFrom tlTo && length lsFrom == length lsTo && and [matchType t1 t2 | (t1,t2) <- zip lsFrom lsTo])
+       else if (matchType tlFrom tlTo && length lsFrom == length lsTo && and [matchType t1 t2 | (t1,t2) <- zip (map snd lsFrom) (map snd lsTo)])
         then -- same handled effects, just use expr
              trace "  same handled effects, leave as is" $
              expr
@@ -160,14 +160,13 @@ resOpen (Env penv gamma) eopen effFrom effTo tpFrom tpTo@(TFun targs _ tres) exp
                          App (makeTypeApp openExpr (map snd targs ++ [tres,effFrom,effTo]))
                              (evExprs ++ [exprVar] ++ [Var p InfoNone | p <- params])
 
-
-                 evIndexOf l
+                 evIndexOf (i, l)
                    = let (htagTp,hndTp)
                              = let (name,_,tpArgs) = labelNameEx l
                                    hndCon = TCon (TypeCon name -- (toHandlerName name)
                                                           (kindFunN (map getKind tpArgs ++ [kindEffect,kindStar]) kindStar))
                                in (makeTypeApp (resolve (toEffectTagName name)) tpArgs, typeApp hndCon tpArgs)
-                     in App (makeTypeApp (resolve nameEvvIndex) [effTo,hndTp]) [htagTp]
+                     in App (makeTypeApp (resolve nameEvvIndex) [effTo,hndTp]) [makeInt32 (fromIntegral i), htagTp]
              in case lsFrom of
                  []  -> -- no handled effect, use cast
                         case lsTo of
@@ -183,7 +182,7 @@ resOpen (Env penv gamma) eopen effFrom effTo tpFrom tpTo@(TFun targs _ tres) exp
                                  else wrapperThunk (resolve (nameOpenNone 0)) []
 
                  [l] -> -- just one: used open-atN for efficiency
-                        trace ("  one handled effect; use at: " ++ show (ppType penv l)) $
+                        trace ("  one handled effect; use at: " ++ show (ppType penv (snd l))) $
                         if (n <= 4)
                          then wrapper (resolve (nameOpenAt n)) [evIndexOf l]
                          else wrapperThunk (resolve (nameOpenAt 0)) [evIndexOf l]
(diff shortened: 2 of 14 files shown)


src/Type/Type.hs
```diff
@@ -559,19 +559,25 @@ makeValueOperation eff tp
...
+    in -- trace ("sorted: " ++ show (map labelName labss) ++ " to " ++ show (map labelName slabs) ++ " and numbered " ++ (show (map (\(i, l) -> show i ++ " " ++ show (labelName l)) (numbered [] slabs)))) $
+       (numbered [] slabs,tl)
   where
+    numbered :: [(Int,Tau)] -> [Tau] -> [(Int,Tau)]
+    numbered [] (l: ls) = numbered [(0,l)] ls
+    numbered ((i, l1):acc) (l2: labs) = 
+        if labelName l1 == labelName l2 
+        then numbered ((i+1,l2):(i, l1):acc) labs
+        else numbered ((0,l2):(i, l1):acc) labs
+    numbered acc [] = reverse acc
```
TimWhiting, line 579: Here is where I number the duplicate labels (they are presorted, so labels with the same name will already be next to each other).

src/Type/Infer.hs
```diff
@@ -1009,10 +1010,10 @@ inferExpr propagated expect (Inject label expr behind rng)
                  -- general handled effects use "@inject-effect"
                  Just coreHTag
                    -> do (maskQName,maskTp,maskInfo) <- resolveFunName nameMaskAt (CtxFunArgs 3 [] Nothing) rng rng
-                         (evvIndexQName,evvIndexTp,evvIndexInfo) <- resolveFunName nameEvvIndex (CtxFunArgs 1 [] Nothing) rng rng
+                         (evvIndexQName,evvIndexTp,evvIndexInfo) <- resolveFunName nameEvvIndex (CtxFunArgs 2 [] Nothing) rng rng
                          let coreMask = coreExprFromNameInfo maskQName maskInfo
                              coreIndex= Core.App (Core.TypeApp (coreExprFromNameInfo evvIndexQName evvIndexInfo) [effTo])
-                                                 [coreHTag]
+                                                 [(makeInt32 0), coreHTag]
```
TimWhiting, line 1018: Not sure about this one, this is in inference of Inject creating the core statement `@inject-effect`. I assumed the new effect will be the first one of that tag, but I'm not sure if I understand this part of inference that well currently.

src/Core/Simplify.hs
```diff
@@ -626,19 +626,18 @@ effectLabelFromHandler tp
...
 effectOffset l effTp
   = let (ls,_) = extractHandledEffect effTp
         ofs = findMatch 0 l ls
     in -- trace ("found offset " ++ show ofs ++ " for " ++ show l ++ " in " ++ show (map (show . labelName) ls)) $
        ofs
 
-findMatch :: Integer -> Name -> [Type] -> Integer
-findMatch i lname (l:ls)  = if (labelName l == lname) then i else findMatch (i+1) lname ls
+findMatch :: Integer -> (Int, Name) -> [(Int, Type)] -> Integer
+findMatch i (off, lname) ((ol,l):ls)  = if (labelName l == lname) && off == ol then i else findMatch (i+1) (off, lname) ls
```
TimWhiting, line 637: This changes to find the label with the correct name at the correct count of the label name.

src/Core/Simplify.hs
```diff
@@ -441,8 +441,8 @@ bottomUp expr@(Case scruts branches)
 
 
 -- simplify evv-index(l) to i if l has a known offset
-bottomUp (App (TypeApp (Var evvIndex _) [effTp,hndTp]) [htag]) | getName evvIndex == nameEvvIndex && isEffectFixed effTp
-  = makeEvIndex (effectOffset (effectLabelFromHandler hndTp) effTp)
+bottomUp (App (TypeApp (Var evvIndex _) [effTp,hndTp]) [App _ [Lit (LitInt i)], htag]) | getName evvIndex == nameEvvIndex && isEffectFixed effTp
+  = makeEvIndex (effectOffset ((fromIntegral i), (effectLabelFromHandler hndTp)) effTp)
```
TimWhiting, line 445: Of course simplification has to take into account the new parameter

src/Core/OpenResolve.hs
```diff
@@ -160,14 +160,13 @@ resOpen (Env penv gamma) eopen effFrom effTo tpFrom tpTo@(TFun targs _ tres) exp
...
-
-                 evIndexOf l
+                 evIndexOf (i, l)
                    = let (htagTp,hndTp)
                              = let (name,_,tpArgs) = labelNameEx l
                                    hndCon = TCon (TypeCon name -- (toHandlerName name)
                                                           (kindFunN (map getKind tpArgs ++ [kindEffect,kindStar]) kindStar))
                                in (makeTypeApp (resolve (toEffectTagName name)) tpArgs, typeApp hndCon tpArgs)
-                     in App (makeTypeApp (resolve nameEvvIndex) [effTo,hndTp]) [htagTp]
+                     in App (makeTypeApp (resolve nameEvvIndex) [effTo,hndTp]) [makeInt32 (fromIntegral i), htagTp]
```
TimWhiting, line 169: This is the only change to open resolve. Adding in the name's count.

lib/std/core/inline/hnd.c
```diff
@@ -73,14 +73,18 @@ static kk_std_core_hnd__ev* kk_evv_as_vec(kk_evv_t evv, kk_ssize_t* len, kk_std_
...
   kk_ssize_t len;
+  kk_ssize_t off = index;
   kk_std_core_hnd__ev single;
   kk_std_core_hnd__ev* vec = kk_evv_as_vec(ctx->evv,&len,&single,ctx);
   for(kk_ssize_t i = 0; i < len; i++) {
     struct kk_std_core_hnd_Ev* ev = kk_std_core_hnd__as_Ev(vec[i],ctx);
-    if (kk_string_cmp_borrow(htag.tagname,ev->htag.tagname,ctx) <= 0) return i; // break on insertion point
+    if (kk_string_cmp_borrow(htag.tagname,ev->htag.tagname,ctx) <= 0) {
+      if (off == 0) return i; // break on insertion point
+      off--; // adjust for duplicate label
```
TimWhiting, line 86: Here is the main change in the runtime. We get the handler with the correct name at the correct offset.

src/Type/Infer.hs
```diff
@@ -540,7 +541,7 @@ inferIsolated contextRange range body inf
          Nothing   -> return res
          Just vrng -> do seff <- subst ieff
                          let (ls,tl) = extractOrderedEffect seff
-                         case filter (\l -> labelName l == nameTpLocal) ls of
+                         case filter (\l -> labelName l == nameTpLocal) (map snd ls) of
```
TimWhiting, line 546: All other changes just look like this essentially (getting the result of `extractOrderedEffect` as it was previously.

daanx:
Thanks so much Tim! 
This is an excellent in-depth investigation -- this is such a hard bug and thank you for pointing out the root cause.
Now, as shown in the [ICFP'21](https://www.microsoft.com/en-us/research/uploads/prod/2021/08/genev-icfp21.pdf) paper, there should be no need for `evv-index` to take another parameter so I thought more on it. I think the error is that the `open` function takes an array of indices and applies them at-once. That is wrong -- if we would apply one at-a-time all indices would be adjusted all the time and `<abort,abort>` would get the right indices `<0,1>`. I adjust for this now in the open resolving adding the masking level to the raw ev index. The commit is 81a8af2 .

Thanks again, Best,  Daan

PR #525, opened 2024-05-19
more strict inference

TimWhiting:
Sorry for all of the whitespace changes in TypeVar.hs due to autoformatting.
The only actual change in that file is changing the import of `Data.Map` to `Data.Map.Strict`.

These changes reduce compilation time from minutes & using up all my RAM to seconds and using 2GB RAM for a simple JSON lexer generated via my koka backend for Alex.


daanx:
That's awesome, thanks Tim!

PR #533, opened 2024-05-22
Fix trailing lambdas getting inserted at the end of the argument list

TimWhiting:
Alternative to #491 

Puts the trailing lambdas before named arguments, but doesn't change anything else.


daanx:
Thanks so much Tim!

PR #549, opened 2024-06-08
Create an overload for function fields to directly call the function

TimWhiting:
Generates another function to access and directly call a function field of a type.

This gets rid of the weird `mystruct.func()()` when getting and then directly calling a function.

I still generate the plain accessors with a locally qualified prefix `thetype/get/function-field`, and generate the caller with the locally qualified prefix `typetype/call/function-field`, this way for no-arg functions you can use `mystruct.call/func()` when the result type doesn't disambiguate.

### DIFF
--- a/lib/std/time/instant.kk
+++ b/lib/std/time/instant.kk
@@ -105,15 +105,15 @@ fun convert( t : timestamp, from : timescale, to : timescale ) : timestamp
   else
     // transform through TAI
     // trace( "convert: " + from.name + " -> " + to.name +  ", t: " + t.show )
-    (to.from-tai)( (from.to-tai)(t) )
+    to.call/from-tai( from.call/to-tai(t) )
 
 
 // Return the `:duration` since the `epoch` for a timestamp `t` interpreted in time scale `ts`.
-pub fun to-tai(  ts : timescale, t : timestamp ) : duration
+pub fun pub/to-tai(  ts : timescale, t : timestamp ) : duration
   t.convert(ts,ts-tai).unsafe-duration
 
 // Given a `:duration` since the `epoch`, return a `:timespan` for that instant in time scale `ts`.
-pub fun from-tai(ts : timescale, d : duration ) : timestamp
+pub fun pub/from-tai(ts : timescale, d : duration ) : timestamp
   d.timestamp.convert(ts-tai,ts)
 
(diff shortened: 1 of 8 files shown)


chtenb:
Is there an alternative solution for this, or will the `()()` syntax remain the intended way of doing this?

TimWhiting:
Daan recently added [this](https://github.com/koka-lang/koka/commit/e307534182482e8c9f6dce4d22a1478d4a8aca6d) on the dev branch. 

So instead of `mystruct.func()()` you can do `mystruct.func.()`. It makes it a bit more clear to the reader what is going on than my solution I think, in addition to the fact that I think my solution probably would result in more inference ambiguities which might confuse people. We probably need to still improve error messages to hint towards the new syntax though if we can detect when it should be added. But the type inference hints need quite a few improvements that we should probably gather under some issue to understand the scope of what we need to improve - especially if it means that we need to do a larger refactor on type inference.

daanx:
@Thanks @chtenb , @TimWhiting , There are some tradeoff's here.  Suppose we have a struct like:
```koka
struct foo 
  v : int
  f : (int) -> int
  g : () -> int

val x = Foo(1, fn(x) x+1, fn() 2)
```

1. With the new `.()` syntax, we can write `x.f.(42)` instead of `(x.f)(42)` which is slightly nicer (or `x.g.()`). Still it might be a bit confusing syntax wise?

2. With Tim's earlier approach of generating overloads for function fields that call it directly, we can write `x.f(42)` to call it, but still use `x.f` to select the function -- very nice!. However, it is problematic for unit functions and `x.g` or `x.g()` both expard to `g(x)` and it is ambiguous whether we want to call the selected `g` field or only select it. 

3. Another solution to consider is to always use "call" overloading for selection of function fields. In such case both `x.g` and `x.g()` call `g`  (and we do not have a "select only" for function fields). That might work well, but in my opinion it still looks a bit strange for `x.g` -- it will select _and call_ `g` without parenthesis. It seems to imply to only select it (like `x.v`) but it may actually have a side effect which is not ideal either.  (note that this is in essence already the case though, we can write `2.println` without parenthesis either so maybe it's ok?)

So, I think we should go with the applier syntax for now (and write `x.g.()` or `(x.g)()` ).

Issue #544, opened 2024-06-02
Array programming


patrick-nicodemus:
It explains in the README that a goal for Koka is to support array programming better in the future, and I wanted to ask what that might look like, and what is currently possible.

1. I want to map a function `f: 'a -> 'a` along a vector, and have this update in place. Is this: currently achievable, an eventual design goal, probably never going to happen, etc.
2. I want to sort a vector of floating point numbers in place: is this ever going to happen?
3. I want to sort a vector of floating point numbers, not necessarily in place but by an algorithm that does not involve converting it into a list, sorting the list, and then converting the list to an array, because this seems to involve much more pointer chasing than a simple ordinary array sort and is probably much slower because of cache locality issues (although I would be very interested to hear if I'm wrong about this)

A related question I have is that, by reading your group's papers, it seems that you can use effects to control stateful computation in the same way that Haskell does with the state monad and runST.

The vectors defined in the standard library are immutable. It is not clear to me whether Koka gives the programmer a way to use scoped effects to do stateful array computations, is there anything there? The documentation for the modify function talks about in-place updates for vectors but I unfortunately did not really understand that bit.

anfelor:
Hi @patrick-nicodemus!

 1. Is [`fun map( v : vector<a>, f : a -> e b) : vector<b>`](https://koka-lang.github.io/koka/doc/std_core_vector.html#map) what you had in mind? This is in-place as long as the input vector `v` is unique.
 2. Sure, this sounds like you want an in-place quicksort?
 3. Sure, there are a few algorithms for this.

> The vectors defined in the standard library are immutable

Actually, they have immutable _semantics_, but they do update in-place if the vector is unique.
You can implement both (2) and (3) in plain Koka, making use of the destructive vector assignment:

```
vec[idx] := value
```

sets index `idx` of `vec` to be `value`. This changes the vector pointed to by `vec`, but it does not change any other vectors in your program. We ensure this by checking at runtime that the memory underlying `vec` is only used here: in that case we can mutate it in-place. Otherwise, we first copy the memory to ensure that no accidental mutation of _other_ vectors occurs.

I don't think that there is any issue with our API for vectors and it should be able to do everything you want to do. However, performance might not be competitive with C++ (but likely still be good enough for your needs). We want to improve two performance aspects in the future:

[...]

PR #550, opened 2024-06-09
Unique vectors

TimWhiting:
Addresses #544 

@anfelor can you take a look at this?

I added a few public apis and a few private helpers.
Public:
- `set`
- `update`
- `update` for local variable vectors

The most important function is the `copy` function, which ensures that the vector is unique prior to applying any mutation operations. For many cases we would expect it to be unique.

I've updated this to use PR #716 here: https://github.com/TimWhiting/koka/commit/b0ef63bfd2241842acc95e34bc5d72c6eae979da


TimWhiting:
We might want to add a few more apis (cutting an array short and dropping the the rest of the array), or extending an array.

lib/std/core/inline/vector.h
```diff
@@ -21,7 +21,6 @@ static inline kk_unit_t kk_vector_unsafe_assign( kk_vector_t v, kk_ssize_t i, kk
   kk_box_t* p = kk_vector_buf_borrow(v,&len,ctx);
   kk_assert(i < len);
   p[i] = x;
-  kk_vector_drop(v,ctx); // TODO: use borrowing
```
anfelor, line 24: Can you rename the `kk_vector_unsafe_assign` function by appending `_borrow` to the name? This signifies that one of the arguments is borrowed.

doc/spec/tour.kk.md
```diff
@@ -647,6 +647,56 @@ that are not possible for general mutable reference cells.
 [Read more about state and multiple resumptions &adown;][#sec-multi-resume]
 {.learn}
 
+### Local Mutable Vectors {#sec-vectors;}
+
+Koka provides a built-in `vector` type that is backed by a constant sized C array of boxed values.
+
+Vectors still provide an immutable interface, but are faster that lists for random access.
```
anfelor: s/that/than/

lib/std/core/inline/vector.h
```diff
@@ -30,3 +29,11 @@ static inline kk_box_t kk_vector_at_int_borrow( kk_vector_t v, kk_integer_t n, k
   kk_box_t b = kk_vector_at_borrow(v,kk_integer_clamp_ssize_t_borrow(n,ctx),ctx);
   return b;
 }
+
+static inline kk_box_t kk_vector_at_own(const kk_vector_t v, kk_ssize_t i, kk_context_t* ctx) {
```
anfelor: Ditto, this should have `_borrow` in its name. I would call the function `kk_vector_unsafe_extract_borrow`.

doc/spec/tour.kk.md
```diff
@@ -647,6 +647,56 @@ that are not possible for general mutable reference cells.
...
+When using vectors locally you can 'mutate' them similarly to local variables:
+
+```
+fun vectors()
+  var myvec := vector-init(10, fn(i) i)
+  myvec[5] := 42
+  myvec.map(show).join(", ").println
+```
+
+When keeping a reference to a vector, the vector will be copied on assignment, but the items in the array will be shared. In the following example, the `copy` will be a vector that shares the same values as `myvec` except at location `5`.
```
anfelor: "If there is another reference to the vector, the vector will be copied on assignment to form the new vector. The elements that were not updated are shared between the old and the new vector."

doc/spec/tour.kk.md
```diff
@@ -647,6 +647,56 @@ that are not possible for general mutable reference cells.
...
+
+```
+fun vectors()
+  var myvec := vector-init(10, fn(i) i)
+  val copy = myvec
+  myvec[5] := 42
+```
+
+When updating a single item in the vector, the old vector will still contain a reference to the value, this will prevent reusing the memory of the old item.
```
anfelor: This sentence is unmotivated. I would write it as: "Koka can even reuse the memory of the elements of the vector if there is only one reference to the element. However, when using the update syntax above, Koka does not release the element from the vector in time for the in-place reuse to occur. If you want to update an element with the opportunity for in-place reuse, you can use the `update` function:" (cross-link the section on reuse, cut the first `reuse()` function)

lib/std/core/vector.kk
```diff
@@ -29,24 +29,62 @@ inline extern unsafe-idx( ^v : vector<a>, index : ssize_t ) : total a
   cs inline "(#1)[#2]"
   js inline "(#1)[#2]"
 
-inline extern unsafe-assign : forall<a> ( v : vector<a>, i : ssize_t, x : a ) -> total ()
+// Obtains the value transferring ownership and replaces the value with uninitialized value
+// Precondition: `v` is unique
+inline fip extern unsafe-idx-own( ^v : vector<a>, index : ssize_t ) : a
```
anfelor: `unsafe-extract`

lib/std/core/vector.kk
```diff
@@ -29,24 +29,62 @@ inline extern unsafe-idx( ^v : vector<a>, index : ssize_t ) : total a
...
   js inline "Array(#1)"
 
+// This is unsafe, since the behavior that depends on this check needs to be observationally equivalent
+inline fip extern unsafe-is-unique( ^v : vector<a> ) : bool
+  c inline "kk_datatype_is_unique(#1, kk_context())"
+  js inline "false"
+  cs inline "false"
+
+// This is technically not fip, but we pretend it to be, since it will only copy when non-unique
+inline fip fun copy(v: vector<a>): vector<a>
```
anfelor: In light of the comments below, a better name might be `ensure-unique`

lib/std/core/vector.kk
```diff
@@ -121,11 +184,44 @@ pub fun foreach-while( v : vector<a>, f : a -> e maybe<b> ) : e maybe<b>
     f(v.unsafe-idx(i))
 
 // Apply a function `f` to each element in a vector `v`
-pub fun map( v : vector<a>, f : a -> e b ) : e vector<b>
-  val w = unsafe-vector(v.length.ssize_t)
-  v.foreach-indexedz fn(i,x)
-    unsafe-assign(w,i,f(x))
-  w
+pub fip fun map( v : vector<a>, ^f : a -> e b ) : e vector<b>
+  // Copy will only happen on the first iteration in the map, to satisfy the unique precondition
```
anfelor: I don't understand this comment: With non-linear control, a copy could happen each iteration?

lib/std/core/vector.kk
```diff
@@ -29,24 +29,62 @@ inline extern unsafe-idx( ^v : vector<a>, index : ssize_t ) : total a
...
+// var v[0] := 1 // updates v in place
+// ```
+// TODO: Is this safe in cs / js? I think we might need to conservatively copy?
+pub inline extern assign/@index( ^self : local-var<s,vector<a>>, ^index : int, assigned : a ) : <local<s>,exn> ()
   c  "kk_ref_vector_assign_borrow"
   cs inline "(#1)[(int)#2] = #3"  // check bounds?
   js inline "(#1)[#2] = #3"       // todo: check bounds!
 
+// Assign to an entry in a local `:vector` variable.
+pub inline fun assign/update( ^self : local-var<s,vector<a>>, ^index : int, f : a -> <local<s>,exn,div|e> a ) : <local<s>,exn,div|e> ()
```
anfelor: Does this require `self` to be unique?

anfelor (changes requested):
Thanks @TimWhiting, this looks a lot friendlier than my draft! I made a few minor comments regarding naming of variables and documentation.

I worry a bit that the unsafe surface area is quite large still. I think we can probably pay for two `is-unique` checks per iteration of `map` (given that the second can be well-predicted). Then we can make this API more clean by incorporating the `copy` calls into `unsafe-assign` and `unsafe-idx-own`.

anfelor:
> We might want to add a few more apis (cutting an array short and dropping the the rest of the array), or extending an array.

Yes, that would be fun! I think we could also support `vector.filter`, which would require us to write `box_null` into the tail of the vector and keep a separate length field. In OCaml, [this is unsupported since `box_null` might not be an inhabitant of the type of the elements though](https://ocaml.org/manual/5.2/api/Dynarray.html#memory_layout), and I wonder if there is a problem that I don't see currently.

lib/std/core/vector.kk
TimWhiting: Leftover from before I realized that I needed an ensure-unique after the call to `f` removed it now.

doc/spec/tour.kk.md
TimWhiting: Thanks, this reads a lot better.

lib/std/core/vector.kk
TimWhiting: Yes, thanks for catching that!

doc/spec/tour.kk.md
TimWhiting: Thanks!

TimWhiting:
> > We might want to add a few more apis (cutting an array short and dropping the the rest of the array), or extending an array.
> 
> Yes, that would be fun! I think we could also support `vector.filter`, which would require us to write `box_null` into the tail of the vector and keep a separate length field. In OCaml, [this is unsupported since `box_null` might not be an inhabitant of the type of the elements though](https://ocaml.org/manual/5.2/api/Dynarray.html#memory_layout), and I wonder if there is a problem that I don't see currently.

I think these sorts of APIs are maybe best left to a resizeable vector type, which could hold on to the length like you mention and have a resize policy for add / extending the vector as well. What do you think?

anfelor (approved):
lgtm

TimWhiting:
I've found a bug with this. Specifically, it behaves poorly with exceptions. Note that this doesn't affect all of the other vector functions, just the mutation of vector elements in local variables (i.e. `assign/update`). Added a few tests that show the problem.

I think we can still do efficient update of mutable variables, but it will require some runtime changes. 
Proposed runtime adjustment: have another special refcount (like STICKY), for REUSE_NO_DROP. For updates to variables, the ref that contains the datastructure needs to be kept alive until the ref is deallocated, but the datastructure itself should have a unique refcount and be reused if it can. However, it shouldn't deallocate itself, if for example an exception is thrown, since the old value could still be obtained via the ref, even if the original value was consumed in the update prior to or after the exception was thrown. 

Turning this to draft for now, until we can discuss this, though I think we should probably move ahead with the other changes which are independent from the local variable change.

TimWhiting:
Superceeded by #716 and #762

PR #560, opened 2024-06-22
fix recursive effects #483

### LINKED ISSUE #483: Internal error when using recursive effect
The following program, when compiled with koka 3.1.1, produces an internal error
```
rec effect process
  fun yield(): ()
  fun fork(p: () -> <process> ()): ()
```

```
> koka -O3 -v0 -o main main.kk
main(1, 1): internal error: label has not kind X: (TCon (TypeCon {typeconName = main/process, typeconKind = KApp (KApp (KCon (->)) (KCon E)) (KApp (KApp (KCon (->)) (KCon V)) (KCon V))}),TCon (TypeCon {typeconName = std/core/types/total, typeconKind = KCon E}))
CallStack (from HasCallStack):
  error, called at src/Common/Failure.hs:46:12 in koka-3.1.1-BaWBegw2eC62R9E4dpMHh4:Common.Failure
  raise, called at src/Common/Failure.hs:32:5 in koka-3.1.1-BaWBegw2eC62R9E4dpMHh4:Common.Failure
  failure, called at src/Common/Failure.hs:28:11 in koka-3.1.1-BaWBegw2eC62R9E4dpMHh4:Common.Failure
  assertion, called at src/Type/Type.hs:744:5 in koka-3.1.1-BaWBegw2eC62R9E4dpMHh4:Type.Type
  appEffectExtend, called at src/Type/Type.hs:679:14 in koka-3.1.1-BaWBegw2eC62R9E4dpMHh4:Type.Type
  effectExtend, called at src/Type/Type.hs:562:14 in koka-3.1.1-BaWBegw2eC62R9E4dpMHh4:Type.Type

Failed to compile main.kk
```

TimWhiting:
Fixes #483

DataInfo is requested to get dataEffect in kind inference, to make a handled effect type with the correct linearity / named information.

However, for recursive effect type declarations we don't have the dataInfo yet for the type in the definition since we are in the process of inferring it. So we need to add the dataEffects for the mutually recursive type definition group prior to inferring them. This makes that change.

### DIFF
--- a/src/Kind/Infer.hs
+++ b/src/Kind/Infer.hs
@@ -372,12 +372,18 @@ bindTypeDef :: TypeDef UserType UserType UserKind -> KInfer (TypeBinder InfKind,
 bindTypeDef tdef -- extension
   = do (TypeBinder name kind rngName rng) <- bindTypeBinder (typeDefBinder tdef)
        qname <- if isExtend then return name else qualifyDef name
+       addRecursiveEffectData qname
        return (TypeBinder qname kind rngName rng, not isExtend)
   where
     isExtend =
       case tdef of
         (DataType newtp args constructors range vis sort ddef dataEff isExtend doc) -> isExtend
         _ -> False
+    addRecursiveEffectData name = do
+      case tdef of
+        (DataType newtp args constructors range vis sort ddef dataEff isExtend doc) ->
+           addEffectData name dataEff
+        _ -> return ()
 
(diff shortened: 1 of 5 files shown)


daanx:
Hi Tim -- thanks for the fix! Ah, I wish I would have looked earlier and more carefully :-(  I just fixed this yesterday when fixing retractive effects and it took awhile debugging this :-(   My fix is commit f28c9fd  and very similar to what you did here. 
Apologies again! I will try to be more on top of the PR's since this is bad.

PR #575, opened 2024-08-28
Variable _host should be exported

hubertwas:
`_host` is a global variable that's used for determining the platform from JavaScript code. It's used from other JavaScript files besides `console.js` and therefore should be exported. I found this trying when trying to execute `test/async/async1.kk` on @TimWhiting's fork with `async.kk`.


daanx:
Thanks!

PR #580, opened 2024-10-04
Add support to read lines longer than 1023 bytes

arvidjonasson:
This pull request includes changes to improve the handling of empty byte sequences in `kk_bytes_cat` and the reading of lines from standard input in `kk_os_read_line`.

### Enhancements to `kk_os_read_line` function:
* Added support to read lines longer than 1023 bytes. Achieved by concatenating 1024-byte chunks until new-line or EOF is reached.
* Fixes problem #571.
* Remove redundant call to `strlen` by allocating string with `kk_string_alloc_from_qutf8n` instead of `kk_string_alloc_from_qutf8`.
* Other small improvements.
(`kklib/src/os.c`)

### Improvements to `kk_bytes_cat` function:
* Added checks to handle cases where either of the byte sequences is empty, such that the function returns the non-empty sequence and drops the empty one. Leads to an early return in these cases.
(`kklib/src/bytes.c`)

### Considerations
* Runtime for `kk_os_read_line` can degrade to `O(n^2)` for **extremely** long lines. However, this seems very unlikely, and I believe this to be a lesser issue than lines being incomplete for lengths `>1023` bytes. To solve this we would have to scale `buf` by a factor `>1` for each iteration in the loop, which can not be done safely on the stack. Thus it's hard to find a balance between solution complexity, constant overhead and algorithmic complexity. My approach prioritizes solution complexity and constant overhead as it stays true to the previous approach. An approach prioritizing solution complexity and algorithmic complexity can be achieved using the `POSIX`/[Dynamic Memory TR](https://en.cppreference.com/w/c/experimental/dynamic/getline) `getline(3)` function.

### DIFF
--- a/kklib/src/os.c
+++ b/kklib/src/os.c
@@ -246,15 +246,36 @@ kk_decl_export int kk_os_write_text_file(kk_string_t path, kk_string_t content, 
 
 kk_decl_export int kk_os_read_line(kk_string_t* result, kk_context_t* ctx)
 {
+  kk_string_t line = kk_string_empty();
   char buf[1024];
-  if (fgets(buf, 1023, stdin) == NULL) return errno;
-  buf[1023] = 0;      // ensure zero termination
-  const size_t len = strlen(buf);
-  if (len > 0 && buf[len-1] == '\n') {
-    buf[len-1] = 0;   // remove possible ending newline character
+  bool newline_found = false;
+
+  // read until '\n', EOF, or error
+  while (!newline_found && fgets(buf, sizeof buf, stdin) != NULL) {
+    kk_ssize_t len = kk_sstrlen(buf); // assumes no embedded '\0' characters
+
+    // check if we have a reached a newline character
+    if (len > 0 && buf[len-1] == '\n') {
+      newline_found = true;
+      buf[--len] = '\0';  // remove newline
+    }
+
+    // append buf to line
+    kk_string_t buf_str = kk_string_alloc_from_qutf8n(len, buf, ctx);
+    line = kk_string_cat(line, buf_str, ctx);
   }
-  *result = kk_string_alloc_from_qutf8(buf, ctx);
-  return 0;
+
+  // success if we reach '\n', or EOF with a non-empty line
+  // this mimics the behaviour of `fgets`
+  if(newline_found || (feof(stdin) && !kk_string_is_empty_borrow(line, ctx))) {
+    *result = line;
+    return 0;
+  }
+  else {
+    // error or EOF with an empty line
+    kk_string_drop(line, ctx);
+    return ferror(stdin) ? errno : EOF;
+  }
 }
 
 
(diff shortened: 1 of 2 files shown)


kklib/src/os.c
```diff
@@ -246,15 +246,29 @@ kk_decl_export int kk_os_write_text_file(kk_string_t path, kk_string_t content,
...
-  const size_t len = strlen(buf);
-  if (len > 0 && buf[len-1] == '\n') {
-    buf[len-1] = 0;   // remove possible ending newline character
+  while (fgets(buf, sizeof buf, stdin) != NULL) {
+    kk_ssize_t len = kk_sstrlen(buf);
+    // check if we have a reached end-of-line
+    if (len > 0 && buf[len-1] == '\n') {
+      buf[--len] = '\0';  // remove ending newline character
+      eol_reached = true; // flag that we reached end-of-line
+    }
```
daanx, line 261: I think we could read zero characters or not get a newline at the end? Maybe it should be something like:
```
eol_reached = (len==0 || len < sizeof(buf) || buf[len-1] == '\n');
```
and keep the check to remove the `\n` but just for that.

kklib/src/os.c
```diff
@@ -246,15 +246,29 @@ kk_decl_export int kk_os_write_text_file(kk_string_t path, kk_string_t content,
...
+    if (eol_reached) break;
   }
-  *result = kk_string_alloc_from_qutf8(buf, ctx);
-  return 0;
+  // success if we reach '\n' or (EOF and we have a non-empty line)
+  if(eol_reached || (feof(stdin) && !kk_string_is_empty_borrow(line, ctx))) {
+    *result = line;
+    return 0;
+  }
+  // error
```
daanx: Make the last case an `else` block?

daanx:
Thanks for the PR -- looks good :-)

kklib/src/os.c
arvidjonasson: I agree with your concerns. My understanding is that `fgets` will always return `NULL` if zero chars are read. So `len` should only be zero if there are`'\0'` chars in the input stream, as `strlen(buf)` will evaluate less than the chars actually read by `fgets`.

However, while thinking about this function I realised I shouldn't have used a `eol_reached` flag at all, since `fgets` also handles some cases where end-of-line is reached (e.g. `EOF` at new iteration).

I have submitted a new version where `fgets` is responsible for `EOF`s and we are responsible for `'\n'`s only. Hopefully this makes the code more descriptive and easier to reason about.

arvidjonasson:
Thanks for the review!

I submitted a new version of `kk_os_read_line`. This should answer your concerns about it and improve the code quality.
Please send suggestions / concerns :)

daanx:
Thanks for the quick update -- I'll take a look soon. These things always turn out to have surprising edge cases or portability concerns.

daanx:
Thanks so much!

PR #595, opened 2024-10-22
fix cpp/java benchmark program: wrong output due to logic error

Guest0x0:
The C++/Java program for the `deriv` benchmark produces wrong output, due to logical error in several pattern match cases. This PR fixes the benchmark programs.

I re-run the benchmark locally, the overall result is unchanged compared to the figures in README.

BTW: I notice that the C++/Java version of `deriv` performs more sharing than the Koka/OCaml version in some cases. For example in this case:
```
    (f,Val(n))             -> add(Val(n),f)
```
the `Val(n)` node is reused directly in C++/Java. This seems a bit unfair? It should not impact benchmark result a lot though, due to Koka's FBIP/OCaml's bump allocation.


daanx:
Very nice! Thanks for fixing this -- and good to hear it didn't invalidate the benchmarks! That would have been a terrible mistake otherwise :-)   All the best, Daan

PR #622, opened 2024-12-18
language server function generation

TimWhiting:
Generates functions for `(==)` `cmp` `show` and `order2` from datatype definition. However, unlike #335 this does it from the language server code action rather than automatically. This allows for more flexibility in working with effects for now, and also allows for greater customization of `show` after generation, while handling most of the boilerplate. We could also add a generic empty template for "adding a function" which would just create a function of one parameter of that type that immediately matches on the datatype, using the correct names of the fields. This would save a lot of typing and make the language server even more useful, while keeping consistent methods using the right / full names of fields.

### DIFF
--- a/koka.cabal
+++ b/koka.cabal
@@ -109,6 +109,7 @@ library
       Syntax.Lexeme
       Syntax.Lexer
       Syntax.Parse
+      Syntax.Pretty
       Syntax.Promote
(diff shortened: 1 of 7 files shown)


woojamon:
+1 I vote for this feature

daanx:
Very cool!

PR #636, opened 2024-12-22
finally get around to fixing issue #9!

### LINKED ISSUE #9: [BUG?] Ambiguous identifier when pattern matching
Consider these two files:

```koka
module foo

import bar

type foo {
  Foo
}

fun to-int(x: foo): int {
  match (x) {
    Foo -> 1
  }
}
```

```koka
module bar

public type bar {
  Foo
}
```

I get this error:

```
src/foo.kk(11, 5): error: identifier Foo is ambiguous. Possible candidates: 
  Foo    : foo
  bar/Foo: bar/bar
  hint   : give a type annotation to the function parameters or arguments
```

This seems wrong: I have specified that `x` has type `foo`, so it isn't ambiguous.

TimWhiting:
Fixes #9

### DIFF
--- a/src/Type/Infer.hs
+++ b/src/Type/Infer.hs
@@ -1672,7 +1672,8 @@ getTypeName tp
 
 inferBranch :: PatternKind -> Maybe (Type,Range) -> Type -> Range -> [Name] -> Branch Type -> Inf ([(Type,Effect)],Core.Branch)
 inferBranch patkind propagated matchType matchRange matchedNames branch@(Branch pattern guards)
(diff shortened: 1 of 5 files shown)


daanx:
Thanks so much Tim! I fixed it based on your commit -- hope that's ok :-)  I use the context to do the search and added a case for 0 parameters for Nil like constructors. Best, Daan

TimWhiting:
Yeah, no problem, looks good! I didn't see that you already had a matchType. Thanks for cleaning it up.

PR #652, opened 2024-12-31
Add some bytes primitives (split and cleaned up from libuv)

TimWhiting:


HeikoRibberink:
I happen to notice that [kk_bytes_set](https://github.com/koka-lang/koka/pull/652/changes#diff-71a4b6b19caa3d07ac28b963652720ce4558767524c6495a1c82dd531e86530eR198) doesn't check the passed index, so [byte/assign](https://github.com/koka-lang/koka/pull/652/changes#diff-8b4bad5e6e5cbd14c2aa55da46baa3456e4784c6023e2592c77098b6afb73d37R69) is unsafe, but this isn't documented.

HeikoRibberink:
This always segfaults when `b` is the empty bytes, probably because `bslice/assign` does not handle this case.
```koka
import std/core/bytes
import std/core/bslice

fun main()
  val b = bytes/alloc(0.ssize_t)
  b.bslice/assign(0, Bslice(bytes/alloc(0.ssize_t), 0, 0, 0))
  ().trace-show
```

TimWhiting:
Sorry, yes, this should be renamed to unsafe-assign or something. In the future I'd love to have an implicit that tries to prove that the index is in range, or checks it and throws otherwise.

HeikoRibberink:
Great! What still surprises me though, is that assigning an empty bslice to an empty bytes causes a segfault. I'd expect this to have no effect, but maybe it's a fault to get the backing buffer of an empty bytes.

TimWhiting:
Yeah, that probably should not segfault. I'll look into it

PR #658, opened 2025-01-11
Add maybe utils

TimWhiting:
A few updates to the standard library for `maybe`. Open to feedback / suggestions.

I've improved the error location for the `unjust` function. I also renamed `unjust2` to `unjust`, since none of the other functions in `core/maybe2` use the `2` qualifier, and we have static overloading which can resolve this most of the time.

I then also added an `expect` function which expects an explicit error message as a second parameter. 
This mimics the naming of this function in Rust's Option type, but I'm open to other names. 

If we go with Rust's naming for this it might be good to look at also changing `unjust` to `unwrap` - maybe with a deprecation notice on the old definition due to it's wide use. Haskell uses `fromMaybe`, Swift uses a postfix `!` and `unsafelyUnwrap` for no runtime checks in non-debug builds.

### DIFF
--- a/lib/std/core/maybe.kk
+++ b/lib/std/core/maybe.kk
@@ -26,10 +26,16 @@ pub fun default( m : maybe<a>, nothing : a ) : a
     Just(x) -> x
 
 // Get the value of the `Just` constructor or raise an exception
-pub fun unjust( m : maybe<a> ) : exn a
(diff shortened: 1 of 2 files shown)


daanx:
Thanks Tim -- I merged for now. Let's discuss potential renaming of maybe to follow Rust/OCaml in person sometime.

PR #661, opened 2025-01-17
fix unique names inside (-) functions in core parsing

TimWhiting:
```koka
fun (-)(l1: list<a>, l2: list<a>, ?(==): (a, a) -> e bool): e list<a>
  l1.filter(fn(x) !l2.any(fn(y) y == x))

fun main()
  [1, 2, 3] - [3, 4, 5]
```

Monadic lifting creates a unique name based on the top level identifier
```koka
// core identifier
fun @mlift-x-@10007[1,0,1,0] : forall<(e :: E)> (@y-x10004 : std/core/types/bool) -> (e :: E) std/core/types/bool;
```
However, `@` is not a valid follow character after `-`

We could change unique names, or change lexing to allow this as an identifier. I chose the latter in this PR.

Also, the implicits translation / open effects can add let identifiers such as the following below:
```koka
fun @lift-handle-cache@10567 // inline size: 3
  = tp... {
    val @implicit/(@uniq-x@10158==)[
```

I edited `Core/Parse.hs` to allow locally qualified identifiers inside function definitions.

### DIFF
--- a/src/Core/Parse.hs
+++ b/src/Core/Parse.hs
@@ -547,7 +547,7 @@ parseDefGroups0 env
 parseDefGroup :: Env -> LexParser (Env,DefGroup)
 parseDefGroup env
   = do (sort,inl,isRec,doc) <- pdefSort
-       (name,_)   <- funid False <|> do{ wildcard; return (nameNil,rangeNull) }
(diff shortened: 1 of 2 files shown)


daanx:
Thanks TIm!

PR #670, opened 2025-02-05
Remove local masks

TimWhiting:
Remove some local masks in the standard library and tests 

I did not change:
- some instances in the v1 std library
- the `misc/` tests (which are not run, and have invalid syntax?)
 - the `type/masklocal*` tests which make sure that you can still have an explicit mask


daanx:
Thanks!

PR #671, opened 2025-02-07
Fix assign/@index JS extern

hubertwas:
I stumbled upon a bug, where I was unable to mutate an array when compile to the `jsnode` backend, but it worked fine in the `C` backend. It turns out, that `local-var`s are wrapped in a map in JavaScript, thus this definition wasn't mutating the vector, and instead was mutating the wrapper. Probably a similar thing is happening in the case of the `cs` inline, but I didn't check.

I tested this with this simple code:
```kk
fun main()
    var a := [0].vector
    a[0] := 1
    println(a[0])
```
using `koka -e --target=jsnode assign.kk`.

It works after this change, and didn't work before.


TimWhiting:
Thanks!

PR #672, opened 2025-02-14
fix macos default triplet for vcpkg

syaiful6:
This PR correct the default vcpkg triplet, in MacOS Koka try to install `pcre2` when loading `std/text/regex` with `vcpkg install pcre2:arm64-macos --disable-metrics`. 

The correct triplet supported is `arm64-osx` and `x64-osx` for MacOS.


daanx:
Thanks!

PR #683, opened 2025-03-31
Add parser combinators `sep-by` and `sep-by1`

toh995:
I wrote this parser combinator for a personal project, and thought it my be useful to add to the standard library!

I took inspiration from haskell, e.g. https://hackage.haskell.org/package/base-4.21.0.0/docs/Text-ParserCombinators-ReadP.html#v:sepBy

### Question:
How do I write tests for the standard library?

Or more generally, is there a "recommended" way to write tests in koka (test framework, etc.)?

I couldn't figure this out, perusing the koka codebase a bit.


daanx:
Thanks!

PR #688, opened 2025-04-06
Properly render `n - 1` in API reference docs

toh995:
Koka's standard library API reference docs were not quite rendering `n-1` properly. I think the fix is to add spaces, i.e. `n - 1`.

image


TimWhiting:
Thanks for the fix!

PR #692, opened 2025-04-13
Add a hint when a variable may be shadowing a function

timbertson:
Adds a hint when a non-callable local variable is used as a function, suggesting that it may be shadowing another function.

I got very confused the first time I encountered the case in the attached test, where a struct accessor (or other "method") gets shadowed by a local variable with the same name. The type error points you in the wrong direction if you don't realise that this shadowing is happening.

I'm _assuming_ that a `Var` whose name is `nonQualified` means "local variable" (or at least global variable in the same file), so is the message accurate here?

## Improvements

I couldn't figure out how to do any of these, but I think they would make this even better:

 - Include the location (filename + range) for where the name (`port` in this test) is defined
 - Print out all the bindings with the same name which the local var is shadowing (i.e. similar to the "ambiguous overload" function listing)
 - (Maybe) only have this hint when you're using dot syntax, i.e. `foo.bar` and not `bar(foo)` since that's the case where this error is the most surprising.

### DIFF
--- a/src/Type/Infer.hs
+++ b/src/Type/Infer.hs
@@ -2440,17 +2440,25 @@ matchFunTypeArgs context fun tp fresolved fixed named
 
     reportNonCallable
       = do
+         hints <- shadowHints
          typeError context range (text "only functions or types with a copy constructor can be applied") tp hints
          return (zip [1..] (map (\x -> ArgExpr x True) (fixed ++ map snd named)), [], typeTotal, typeUnit, Core.App)
       where
-        hints
+        shadowHints
           = case fun of
-              Var name _ nameRange | not (isQualified name) ->
-                [(
-                  text "hint",
-                  text ("`" ++ showPlain name ++ "` at " ++ showRange "." False nameRange ++ " might be shadowing another function")
-                )]
-              _ -> []
+              Var name _ nameRange -> do
+                vals <- lookupLocalName isInfoVal name
+                ppEnv <- getPrettyEnv
+                case vals of 
+                  Right (nm, nameInfo) -> 
+                    return $ shadowHint ppEnv (nm, nameInfo)
+                  Left results -> return (concatMap (shadowHint ppEnv) results) -- Multiple locals with the same name?
+              _ -> return []
+        shadowHint ppEnv (nm, nameInfo) = 
+          [(
+            text "hint",
+            ppName ppEnv nm <+> text "at position" <+> text (show (infoRange nameInfo)) <+> text "might be shadowing another function"
+          )]
 
 
 
--- a/src/Type/Infer.hs
+++ b/src/Type/Infer.hs
@@ -2325,8 +2325,7 @@ matchFunTypeArgs context fun tp fresolved fixed named
                                 = let coreVar = coreExprFromNameInfo qname info
                                   in (Core.App (coreInst coreVar) (seqqList coreArgs))
                           return (iargs,pars,eff,res,coreAddCopy)
-                  _ -> do typeError context range (text "only functions or types with a copy constructor can be applied") tp []
-                          return (zip [1..] (map (\x -> ArgExpr x True) (fixed ++ map snd named)), [], typeTotal, typeUnit, Core.App)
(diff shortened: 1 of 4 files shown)


TimWhiting:
Yeah, it would be best not to assume unqualified names are locals. 

We should use the NameInfo portion of the matches (map snd matches), that will let us know whether it is a value or function we are referring to and since values could be lambdas we probably also want to look at the type scheme in the infoType field to determine whether the val is a function. If any of the matches contains a value instead of a function we can report it, but I wouldn't necessarily indicate that it is a local variable, because it could be a top-level variable as well. We should probably add more info to the NameInfo to indicate whether a value is top-level or not.

timbertson:
> since values could be lambdas we probably also want to look at the type scheme in the infoType field to determine whether the val is a function

The code branch I'm modifying is reporting that the subject isn't callable, so does that mean we can skip this check?

If it's not easy to detect local from toplevel that seems fine (I can just make the hint less specific), but is there a way to insert the location of the variable's definition? That seems like the most useful information.

TimWhiting:
https://github.com/koka-lang/koka/blob/199a5ad9f7924f8c4a6a2e1e6fc23a6c3249cb8a/src/Type/Assumption.hs#L67

There should be source range information from the NameInfo

timbertson:
Is there a file path in there somewhere? The `Range` type only seems to be a file number/column, but what if the var is defined in a different module?

timbertson:
Silly me, I thought `Range` was just the lineno/column and I had to figure out the path myself. I've pushed a commit which includes the definition location, but I did have to pass `"."` as the `cwd` argument, I feel like that's cheating and I should be getting it from somewhere else?

TimWhiting:
If this is a local variable specific hint, I don't think we should be adding the module path to the range. I've pushed an update to your branch which simplifies to just the basic range. Also, you were using the range of the application site. I've updated to get the range for the local variables. Do the changes look good to you?

timbertson:
> If this is a local variable specific hint, I don't think we should be adding the module path to the range

It looks like you've updated this to detect _only_ locals, thanks LGTM 👍 

TBH I'd still like to include the path though, since (in VSCode at least) you can cmd+click any filepath:lineno text to jump to that source location but if it's just the lineno you have to navigate yourself. But it's not a big deal if you would prefer to keep that redundant info out.

TimWhiting:
In the case of VSCode if we format it right we can actually get a hyperlink while still only showing a shorter thing. Probably just hyperlink the name I guess. I can take a look later. However we should only show that in an IDE setting.

timbertson:
Maybe I'm a weirdo, but I usually run the compiler in the terminal and
click on stuff rather than going through the "problems" list from the LSP.
Maybe it's too many bad experiences where the language server is serving up
stale errors.

TimWhiting:
I'm going to merge this as is for now. 

I think we should go through and add range information (and clickable links for editors) to improve all of the error messages, not just this one. 
That will require us to know if we are in an editor context or not, which is beyond the scope of this PR. 

As far as the LSP serving up stale errors, I wonder if you could give some more insight into when this happens so we can fix the root cause. Create an issue if you have some small reproducible steps. I know that I often end up restarting the language server using the VSCode command menu, but sometimes that is because I'm working on different versions of the compiler, language server, and extensions, and play around with more unstable features. Restarting the server should clear your errors unless you are in a state where the compiler dies immediately on restart due to some crash type checking your code. We need to try to fix those crashes I think.

timbertson:
> I'm going to merge this as is for now.

Thanks, I was going to suggest the same :+1:

> As far as the LSP serving up stale errors, I wonder if you could give some more insight into when this happens so we can fix the root cause.

I think this may just be a bit of an OCD / paranoia workflow, and isn't specific to koka. I'm often working with a big scala project, and when there's a lot of errors the LSP can take a while to catch up since it simply takes a while to compile. Trusting it isn't worth my time, as some of the errors will be false positives when it's still catching up. Whereas if I compile in the terminal, then I know all the errors I get in that command's output _definitely_ refer to the latest state.

PR #694, opened 2025-04-17
rename ref.modify -> ref.access

timbertson:
This [caused me some confusion](https://discord.com/channels/1231710450152902716/1231710450152902719/1349688983822536744) when I was getting started, particularly because `ref.modify { x => x + 1 }` is exactly how you would increment a Ref[Int] in scala's cats-effect, but produces surprising results in koka.

When I found out how it worked I thought `access` seems like a more neutral name, since it allows reading & writing, whereas to me `modify` is misleading because I assumed it would _always_ modify the ref.

I've left `modify` as a deprecated alias.


TimWhiting:
I'm a bit worried about the docs for this because the vector use case mentioned isn't really accurate, we need something like #716 or #550 for vectors to be able to update in-place. This only puts the ref into a local scope, which prevents escaping the scope, nothing else as far as I can tell. Maybe a better name for this would be `as-local` or `local-scope` or something?

timbertson:
I think those names work (though `access` still feels accurate to me too).

Regarding the docs, what's the core difference? IIUC it's that using `modify` will not end up creating two references to the value. Which would always save a refcount incr/decr, but if the underlying value is unique (likely) it'd also  enable all the in-place optimisations of modifying a unique reference?

TimWhiting:
Yeah, that is the core difference. Access only makes the `ref` seem like a local (i.e. no escaping the scope and implicit dereferencing). Under the hood it is still just a `ref` value, not the inner value.

PR #698, opened 2025-04-26
colorscheme tweaks to use red sparingly

timbertson:
Very minor issue, but I find the compiler output can be tough to scan as there's a lot of red, and I'm typically trying to focus on errors so ideally they'd be the main red part.

I played around a bit and came to something that I quite like.

Here's a comparison output with both a warning and an error, to illustrate the difference:

Current scheme:
Screenshot 2025-04-26 at 12 41 32 pm

Less reds:
Screenshot 2025-04-26 at 12 38 54 pm

### DIFF
--- a/src/Common/ColorScheme.hs
+++ b/src/Common/ColorScheme.hs
@@ -67,8 +67,8 @@ defaultColorScheme
   = darkColorScheme
 
 darkColorScheme
-  = let c = emptyColorScheme{ colorInterpreter = DarkRed
-                            , colorCommand     = Red
+  = let c = emptyColorScheme{ colorInterpreter = DarkCyan
+                            , colorCommand     = Magenta
                             , colorError       = Red
                             , colorComment     = DarkGreen
                             , colorReserved    = DarkYellow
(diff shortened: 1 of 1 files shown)


chtenb:
Not bad, I agree that red should be used more sparingly. I think it would be nice if we can leave the `check :` stuff highlighted distinctly. I think magenta would be a good choice here.

TimWhiting:
I agree that check : or file name stuff should still be highlighted differently. Possibly with the grey of the default text.

timbertson:
Updated to use magenta @chtenb . I think grey/default might be confused with other unfiltered output, it's good to have it be coloured in some way.

timbertson:
Oh and files in the current directory (or just outside the stdlib? not sure) are already printed in grey, so it'd good to maintain a different colour for the prefix.

TimWhiting:
Can you update the screenshot with the magenta for this PR?

timbertson:
It's green now (https://github.com/koka-lang/koka/commit/0c7d5133eac8447af11fbd5f025c4fdd9c49b6fa) (which I like fwiw):

Screenshot 2025-07-03 at 9 41 54 pm

Issue #702, opened 2025-05-03
Can't exit the process with nonzero status


timbertson:
I couldn't find any function in the stdlib to exit the program, I'd expect e.g. `os/process` to expose an `exit(status: int) -> io a` function.

Given this isn't available, I tried to `throw("error!")` to signal failure but noticed that this results in a successful exit status too:

```
// exit.kk
fun main()
  throw("error!")
```

```sh
$ koka -e exit.kk && echo "OK!"
# ( ... )
uncaught exception: error!
OK!
```

PR #703, opened 2025-05-03
Add exit code to default exception handler

TimWhiting:
Addresses #702

### DIFF
--- a/lib/std/core.kk
+++ b/lib/std/core.kk
@@ -41,6 +41,7 @@ pub import std/core/console
 extern import
   c  file "core/inline/core"
   cs file "core/inline/core.cs"
+  js file "core/inline/core.js"
 
 // ----------------------------------------------------------------------------
 // Builtin effects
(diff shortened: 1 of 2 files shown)


lib/std/core.kk
```diff
@@ -229,14 +229,22 @@ pub extern host() : ndet string
...
   js inline "$std_core_console._host"
 
+// Exits the program with the specified exit code
+// 
+// *Warning:* 
+// * In Javascript this throws non-handleable (by Koka) exceptions even for an exit code of 0.
+// * In C this just terminates the program. Note that you are typically limited to an exit code in an 8-bit range.
+pub extern exit(i: int): io-noexn ()
+  c inline "exit(kk_integer_clamp32(#1, kk_context()))"
+  js inline "throw(`Exited with code: ${#1}`)"
```
timbertson: I hit this today and it didn't quite work, I get:

```
file:///Users/tcuthbertson/Code/scratch/kokatest/koka-std/.koka/v3.1.3/js-debug-70c9ee/std_core_dash_extras.mjs:800
  return throw(`Exited with code: ${i}`);
         ^^^^^

SyntaxError: Unexpected token 'throw'
    at compileSourceTextModule (node:internal/modules/esm/utils:338:16)
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:102:18)
    at #translate (node:internal/modules/esm/loader:468:12)
    at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:515:27)
```

Since it comes after `return` it needs to be an expression, this looks a bit silly but it works:

```suggestion
  js inline "(function(code) { throw(`Exited with code: ${code}`) })(#1)"
```

And then I get:

```
node:internal/modules/run_main:122
    triggerUncaughtException(
    ^
Exited with code: 1
(Use `node --trace-uncaught ...` to show where the exception was thrown)
```

timbertson: Improving on this, an extern js is more readable, plus it can be smarter on nodejs (if you use process.exit then you don't get the useless stacktrace printed to the console):

```js
function kk_exit(code) {
  if (globalThis.process && globalThis.process.exit) {
    process.exit(code) // nodejs
  } else {
    throw new Error(`Exited with code: ${code}`) // browser
  }
}
```

TimWhiting: Thanks! tested manually and works great!

TimWhiting:
@timbertson I just realized this might already be available as `core/debug/impossible`...

timbertson:
That terminates the program unsuccessfully, but I'd still want a plain exit function which doesn't print any extra messages and does let you choose the exit code (that part's rarely used these days, but can't hurt to expose).

PR #712, opened 2025-05-17
os.list-directory: fix handling of errno

timbertson:
I was confused that `list-directory` always seemed to return `[]`.

Digging in, I found that [you need to set `errno` to 0 before calling `readdir`](https://man7.org/linux/man-pages/man3/readdir.3.html), as otherwise you can't tell the difference between the end of the directory and an actual error - you'll just see whatever `errno` happened to be before the call.

For whatever reason errno happened to be 9 (EBADF) from some prior syscall. The loop didn't check for errors when iterating over entries, so I've also added a check for that.

Aside: it'd be nice to have some of [std/data/sort.kk](https://github.com/koka-community/std/blob/04e3726f9aef14409da9541d0ef9342fa31c3276/std/data/sort.kk) in the stdlib, I was hoping to print out the filenames for verification in the test I added but I couldn't sort them.

### DIFF
--- a/kklib/src/os.c
+++ b/kklib/src/os.c
@@ -531,8 +531,12 @@ static kk_string_t os_direntry_name(dir_entry* entry, kk_context_t* ctx) {
 #define dir_cursor DIR*
 #define dir_entry  struct dirent*
 static bool os_findnext(dir_cursor d, dir_entry* entry, int* err) {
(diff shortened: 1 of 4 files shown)


timbertson:
I was also surprised to see that `list-directory` returns `[]` when there's an error - could it raise the error and let higher level code make this decision?

PR #714, opened 2025-05-17
fix: use mainentry flag when evaluating as part of compilation

timbertson:
This logic should probably be put somewhere common, but I wasn't sure where. For now I've copied it from [CodeGen.hs](https://github.com/koka-lang/koka/blob/8d9004e97b00b4831296723164b7bde4e3c92787/src/Compile/CodeGen.hs#L290)


TimWhiting:
@timbertson 
I noticed that this wasn't working to run a series of entry points that I was expecting.

For example say that I wanted to run all the test functions in some files, this wasn't actually running those files when I specified `-e --mainentry=test testfile1 testfile2`. 

This is because the current flag changes the name of the generated target name. (For example, let's say that I want to use some C framework that needs it's own main, and I want to generate a function named `koka_main` to run my Koka code within that framework). So this was generating C files that had their `main` functions renamed to `test`. (Also it was looking for tests to generate).

That seems to be a different use case completely (and something I remember someone requesting). In some cases it might make sense for them both to be the same, but I can see use cases where you want to control them separately.

I've pushed up a change that introduces a new flag `--maintarget` for the target functions you want to compile in those files.

Unfortunately the naming is a bit weird. I think the names should be swapped from what I introduced, but that would break anyone who is relying on the old name. Any suggestions for different names? To be honest I'm not too concerned about breaking anyone at this point, but right now we don't have a good breaking-change process or communication channel.

timbertson:
Ah, thanks for figuring that out.

Agree it's confusing, if we're going to change them perhaps the flag which sets the generated name could be `--output-entrypoint=...` (to match `--output`), and the flag which controls which koka function gets run could be `--entrypoint`? That way nobody would silently be getting the old functionality, they'd start seeing an unsupported option if they were using `--mainentry`.

daanx:
Thanks so much!

PR #738, opened 2025-06-27
Reorganize Version Management of VSCode Extension and add version download from github.

TimWhiting:
I did quite a bit of reorganizing and simplifying in the VSCode extension. 

In particular: 
- I removed parameters that were unused from methods that did not need them
- I moved some platform and path constants into their own file `platform-paths.ts`
- I split the `workspace.ts` file into a `workspace-configuration.ts` (aka kokaConfig) and a `version-manager.ts`
  - Version manager handles discovering installed compilers, managing the compiler path, and checking for and installing / uninstalling versions of the compiler.
    - I moved some top level functions to methods as part of the version manager class, removing parameters that are now accessible via `this`.
  - Workspace Configuration / kokaConfig on the other hand, handles various general configuration options of the workspace - default compiler arguments, workspace / include directories, compiler targets, language server options. Anything that doesn't have to deal with versions of the compiler, (though versionManager is a field of the kokaConfig now). 

These changes introduce a few new features:
- Automatic check for updates of the compiler.
- Installation of prereleases when the setting `koka.dev.usePrereleaseCompilers` is enabled.
- A new command to choose a specific version of Koka to download from all the tagged releases on GitHub.

# TODO
CI: 
- [ ] Update CI to publish the whatsnew file as well

Extension:
- [ ] Add a minimum supported SDK and filter all releases to those greater than the minimum.
- [ ] Fetch the whatsnew file from the release.


TimWhiting:
Pros:
- New extension doesn't need to be published for every compiler update. We could even get `whatsnew.md` from the release on github or include that in the bundle. Most compiler updates shouldn't require us to touch or republish the extension - which means that the release process is entirely automatic via github actions, but still controlled by a human changing the release from draft status to a prerelease or latest release on Github. In general this removes duplicated effort for a release.
- When a bug occurs it is easy to revert to an older version of the compiler.
- It is easy to manually or automatically install pre-release versions of the compiler, for users to test out bug-fixes.
- Most users do not downgrade extensions, and extensions usually are set to automatically be updated. If the compiler breaks it would be unintuitive for a user to have to downgrade the extension especially if they are unaware of that process (which I wasn't). Whereas if there is version management in the extension it would be easy for users to be aware of how to downgrade the compiler.

Cons:
- Older compilers might not work with current version of extension if the extension breaks backwards compatibility. This can be solved by setting a minimum supported compiler version. Fortunately we are using the language server protocol (which only makes backwards-compatible changes). We do have a few custom commands / endpoints such as compiling a certain function, however, it should be very rare that the interface between the compiler and extension changes for these functions. New custom endpoints can be gated by a version they were introduced in and an error message surfaced or the feature disabled if the user is on an older version of the compiler. Only backwards-incompatible changes to the custom endpoints would need to update the minimum compatible compiler version for the extension as a whole, unless the extension adapts - which I suggest it shouldn't to keep things simple.

I feel like there were other cons we discussed, and am open to removing the auto update aspect of this PR.

daanx:
Thanks!

PR #740, opened 2025-06-28
Fix regex-inline for replace-all

ov7a:
A reproducer for the issue:
```
import std/text/regex

fun main()
  trace("some text with o".replace-all(regex("[o]"), "a", 1))
```
```
koka --target=js test-regex-all.kk
```
It will fail with
image
in browser because `_vlist` is defined in a vector:
* https://github.com/koka-lang/koka/blob/561a11cc1944f2cd974562b329c0a31b63365b9e/lib/std/core/inline/vector.js#L10


TimWhiting:
Thanks!

Issue #659, opened 2025-01-12
String functions behavior depens on target (Unicode support)


ov7a:
C target treat strings as ASCII:

https://github.com/koka-lang/koka/blob/a39737aaf12baea8af1354f013487250f6698253/kklib/src/string.c#L831

while JS treats them as unicode:
https://tc39.es/ecma262/multipage/text-processing.html#sec-string.prototype.tolowercase

Simple reproducer:
```
$ echo 'println("Добар Дан".to-lower)' | koka --target c
...
Добар Дан
$ echo 'println("Добар Дан".to-lower)' | koka --target js
...
добар дан
```
`trim` operation is also affected by this. In JS it's Unicode, while in C it's not:
```
$ echo 'println("a\xA0".trim ++ ".")' | koka --target c
...
a .
$ echo 'println("a\xA0".trim ++ ".")' | koka --target js
...
a.
```

As a user, I'd expect that all targets support Unicode.

TimWhiting:
I think Daan has mentioned that he'd like the core c libraries to be minimal, so that we can create really small binaries. 

However, I agree that in most cases you want unicode behaviors. Maybe we could the name of these functions to `ascii-lower`, and `ascii-trim` etc, or just move the definitions to an ascii module so that they are `ascii/lower` and `ascii/trim`, with `unicode/lower` and `unicode/trim` being in the unicode library. Of course we would also need to then add to the Javascript runtime for the ascii library for correct ascii to-lower.

This way for executables that need unicode support we can include the static libraries for unicode support, otherwise we can leave them out. Going the route of including the c libraries for unicode, we'd want them to only linked as part of the unicode library like we do with regex, see: https://github.com/ov7a/koka/blob/3fd97ec53192db120954cc2cdf374a890df8b6c4/lib/std/text/regex.kk#L18

PR #741, opened 2025-06-28
Make JS regex unicode

ov7a:
By default, JavaScript regex are not unicode.

However, C and C# regex are unicode.

To test this, you can run
```koka
import std/text/regex

fun main()
  println("Добар Дан".split(regex("[^\\p{L}]+")).show)
```
With C and C# you'll get 2 words. With JS target
```
koka -e --target=js test-regex2.kk 
```
you'll get
```
["",""]
```

This PR adds unicode flag to JS regex, so the behavior is consistent.

Related:
* https://github.com/koka-lang/koka/issues/659


daanx:
Thank you!

PR #748, opened 2025-07-11
Add eta-expansion and tests

TimWhiting:


daanx:
Thanks Tim !  I will merge but make some changes:
- Let's not do warnings in the parser for outer level _ ; we can do that in the type checker and avoid mapping over user syntax
- We should consider allowing parenthesized eta expansion? like `(inc(_))` at toplevel is `(fn(x) inc(x))` 
- The etaTransform is not a state monad but an output monad -- and I think for such short functions let's just return the binders directly without using a monad

HeikoRibberink:
This is great!
Do I understand correctly that this PR: 
- Introduces a shorthand for `fn(x) <code using x>`, by allowing `_` in expressions, and,
- Changes the hole in constructor contexts from `_` to `hole`?

TimWhiting:
@HeikoRibberink, yes, that is what this is doing. Constructor contexts always have allowed `hole`, but now the `hole` syntax will be required.

Since this is a purely syntactic transformation it introduces the lambda and binders at the *innermost* argument that is not a plain `_`. So it won't allow a shorthand for *any* lambda expression, just simple ones. I believe it will cover most common places that are annoying to create an explicit lambda currently.

PR #752, opened 2025-07-17
fix(doc): correct argument order in zipwith eta-expansion example

amistozy:


TimWhiting:
Thanks! Can't believe I missed that.

PR #754, opened 2025-07-17
Fix named parameters for higher order functions

TimWhiting:
@daanx 

Not sure if we should also adjust https://github.com/koka-lang/koka/blob/c80571c92b7d22b4a3b186617ed64f1952ab711a/src/Type/InferMonad.hs#L1341

It seems like currently we don't take into account the named-everywhere aspect for resolving implicits.

This fails to type check (with a weird error I will note):
```koka
val basic/an-implicit = 3
fun param(a : int = 0, b : int = 0, ?an-implicit: int) : console int
  println(a.show ++ ", " ++ b.show)
  0
fun impl-higher-order(?param : (b: int) -> console int)
  param(b = 1)
fun test()
  impl-higher-order()
```

And this causes type checking to freeze:
```koka
val basic/an-implicit = 3
fun higher-order(f : (b: int) -> console int)
  f(b = 1)
fun param(a, b, ?an-implicit: int) : console int
  println(a.show ++ ", " ++ b.show)
  0
fun impl-higher-order(?param : (b: int, a: int) -> console int)
  param(1, 2)
fun test()
  impl-higher-order()
```


daanx:
Interesting .. but also a bit scary how the optional parameters cause so much edge cases. Let's not forget to address the other bugs/concerns with this but we can do this later on.

TimWhiting:
Sounds good. I've added an issue #755 so we don't forget to address the other problems later.

Issue #510, opened 2024-04-24
Suggestion: `rcontext.resume-shallow` should not run the `return` clauses


anfelor:
Unlike Frank, Koka does not remove the return clause when resuming a shallow handler (this was originally spotted by @jasigal). This is currently necessary since `resume-shallow` has the same return type as `resume` (which runs the potentially-type-changing return clause): https://koka-lang.github.io/koka/doc/std_core_hnd.html#type_space_resume_context

```koka
fun resume( r : resume-context<a,e,e1,b>, x : a ) : e b
fun resume-shallow( r : resume-context<a,e,e1,b>, x : a ) : e1 b
```

However, this is quite annoying in practice. For example, we could attempt to get Frank-style pattern matching on shallow handlers by writing:

```koka
effect send<a>
  ctl send(x : a) : unit

type ssend<a,e,b> { SDone(b : b); Send(x : a, k : () -> <send<a>|e> b) }

fun canthappen() : a
  fun go() { go() }
  unsafe-total(go)

fun ssend(f : () -> <send<a>|e> b) : e ssend<a,e,b>
  with handler
    return(x) { SDone(x) }
    raw ctl send(x)
      Send(x, fn()
        match rcontext.resume-shallow(()) // we would expect to get 'b' but get 'ssend<a,e,b>'
          SDone(y) -> y                   // unwrap the return clause
          Send(_, _) -> canthappen()      // we removed the effect handler so no Send constructor can happen
      )
  f()
```

This is not just a bit ugly, but also less efficient since `resume-shallow(())` should really be a tail-call here.

I would guess that to fix this the `resume-context` should carry two different types: The return type before running the `return()` clause and the type after running the `return()` clause. Then in https://koka-lang.github.io/koka/doc/std_core_hnd-source.html#prompt the case:

```koka
  Shallow x ->
    yield-bind( cont({x}), fn(y) ret(y) )
```

should really be:

```koka
  Shallow x ->
    cont( {x} )
```

TimWhiting:
#825 Shows how much performance we lose due to this. Previously stack space and exponential slowdown, with a hacky fix 2-6 million context switches per second depending on hardware.

jasigal:
Are there any plans to make this change? I love prototyping theory research ideas in Koka (rank-N types, implicits, effect type system, performance), but this has been a stopper in a few cases where I need/really want shallow effects.

[...]

PR #783, opened 2025-08-07
Adjust finalization

TimWhiting:
#780 brought this up. 

The parameter/return value for `rcontext.finalize` does not make much sense in practice. `raw ctl`'s `rcontext` is intended to escape the handler's return context, and the required parameter makes it awkward to implement cooperative asymmetric coroutines. 

The other place `Finalize` is used is to ensure finalization if `ctl` clauses do not resume, in which case the return value is important to keep. Instead of creating a new variant of `FinalizeRaw : resume-context`, this PR just casts the return type to `()`. 

Obviously there are other potential approaches, this probably needs to be discussed more, and a test would need to be added.


TimWhiting:
Related #510

PR #795, opened 2025-08-29
Add $KOKA_PATH, and pkg-config lookup for C libraries

timbertson:
This PR implements the two ideas I mentioned [in discord](https://discord.com/channels/1231710450152902716/1231777420072194068/1410226410529755267):

 - use `$KOKA_PATH` as an additional list of include directories, similar to many other languages
 - add support for resolving C libraries via [`pkg-config`](https://www.freedesktop.org/wiki/Software/pkg-config/). This is a standard which is supported by a number of build tools (including many conan / vcpkg packages). This may make the existing two integrations redundant one day, but for now I've just added it as an additional source.

Sample output finding `libuv` with `--verbose=4` (the library is provided by `nix`, thus the long paths):

```
codegen : uv/utils
generated c: .koka/v3.2.2/clang-debug-5f0454/uv_utils
libPath : libuv.a not found in ["/usr/local/lib","/usr/lib","/lib","/opt/homebrew/lib"]
command : /nix/store/hj7iydwvhkslzsvw82iilv8say0xv5n7-pkg-config-wrapper-0.29.2/bin/pkg-config --variable=libdir libuv
out     : /nix/store/nwck8lyydw82yvxc8gz9h5b92yhwdcy6-libuv-1.50.0/lib
libPath : /nix/store/nwck8lyydw82yvxc8gz9h5b92yhwdcy6-libuv-1.50.0/lib/libuv.a
command : /nix/store/hj7iydwvhkslzsvw82iilv8say0xv5n7-pkg-config-wrapper-0.29.2/bin/pkg-config --variable=includedir libuv
out     : /nix/store/sqbwc4gyayj983i6qkw3bmikxbf86v2c-libuv-1.50.0-dev/include
library : /nix/store/nwck8lyydw82yvxc8gz9h5b92yhwdcy6-libuv-1.50.0/lib/libuv.a
codegen done: uv/utils
link    : uv/utils
command : /nix/store/xpja3ad45p55aryvgng6g9grp8yivvsd-clang-wrapper-19.1.7/bin/clang -Wall -Wextra -Wpointer-arith -Wshadow -Wstrict-aliasing -Wno-unknown-pragmas -Wno-missing-field-initializers -Wno-unused-parameter -Wno-unused-variable -Wno-unused-value -Wno-unused-but-set-variable -Wno-cast-qual -Wno-undef -Wno-reserved-id-macro -Wno-unused-macros -Wno-cast-align -Wno-unknown-warning-option -g -Og -c -I /Users/tcuthbertson/.local/share/koka/v3.2.2/kklib/include -I /nix/store/sqbwc4gyayj983i6qkw3bmikxbf86v2c-libuv-1.50.0-dev/include -DKK_MIMALLOC=8 -o .koka/v3.2.2/clang-debug-5f0454/uv_utils.o .koka/v3.2.2/clang-debug-5f0454/uv_utils.c
building: 0.822507s
```

### DIFF
--- a/src/Compile/Options.hs
+++ b/src/Compile/Options.hs
@@ -862,7 +862,10 @@ processInitialOptions flags0 opts
               (localDir,localLibDir,localShareDir,localBinDir)
                   <- getKokaDirs (localLibDir flags) (localShareDir flags) buildDir
 
-              normalizedIncludes <- mapM realPath ("." : (localShareDir ++ "/lib") : includePath flags)
+              pathsFromEnv <- fmap undelimPaths (getEnvVar "KOKA_PATH")
+              let allIncludes = nub $ filter (not . null) ("." : (localShareDir ++ "/lib") : includePath flags ++ pathsFromEnv)
+              normalizedIncludes <- mapM realPath allIncludes
+              -- _ <- trace ("Normalized includes: " ++ show normalizedIncludes) $ return ()
 
               -- cc
               ccmd <- if (ccompPath flags == "") then detectCC (target flags)
--- a/src/Common/File.hs
+++ b/src/Common/File.hs
@@ -15,7 +15,7 @@ module Common.File(
                   , searchPaths, searchPathsSuffixes, searchPathsEx, searchPathsCanonical
                   , getMaximalPrefixPath
                   , searchProgram
-                  , runSystem, runSystemRaw, runCmd, runCmdRead, runCmdEnv
+                  , runSystem, runSystemRaw, runCmd, runCmdRead, runCmdReadExit, runCmdEnv
                   , getProgramPath
 
                   -- * Strings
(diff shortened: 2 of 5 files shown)


src/Compile/Options.hs
```diff
@@ -853,7 +854,10 @@ processInitialOptions flags0 opts
               (localDir,localLibDir,localShareDir,localBinDir)
                   <- getKokaDirs (localLibDir flags) (localShareDir flags) buildDir
 
-              normalizedIncludes <- mapM realPath ("." : (localShareDir ++ "/lib") : includePath flags)
+              pathsFromEnv <- fmap splitSearchPath (getEnvVar "KOKA_PATH")
```
TimWhiting: It looks like we already have a utility for splitting delimited paths used elsewhere in this file: `undelimPaths`. Not sure if Daan would want to avoid the extra dependency, and use this, or if we should use a version that might handle slightly more edge cases than we do. We tend to avoid adding more dependencies.

timbertson: Thanks, happy to use that - I wasn't aware of it. GHC reported that `filepath` was a hidden dependency which I assumed meant it was already required transitively, but there's a chance it was only used for tests previously.

TimWhiting: Hmm, yes, that might be the case.

PR #798, opened 2025-08-30
fix exit status

### LINKED ISSUE #797: run-system returns invalid codes
```koka
import std/os/process

fun main()
  val status = run-system("exit 3")
  println(status)
```
will return... 768. "exit 1" will return 256.

FTR, 
```
$ sh -c "exit 1"
$ echo $?
1

$ sh -c "exit 3"
$ echo $?
3
```

TimWhiting:
Fix exit status on unix for `system` call. The returned integer is not directly interpretable as an exit status.

Fixes: #797 

Unfortunately, unix does this to require the user to check for other abnormal termination (due to signal / stopped process). 

Not sure how you want to report that?
- Add a tagged C union in kklib, and convert it to a Koka Error type in std/os/process? 
- Assignable parameters in kklib, ditto for std/os/proccess?
- Something else?

My opinion is to do the simple fix now (below), with the plan of re-addressing this when we get to real process integration with libuv.

@daanx

### DIFF
--- a/kklib/src/os.c
+++ b/kklib/src/os.c
@@ -648,7 +648,15 @@ kk_decl_export int kk_os_run_system(kk_string_t cmd, kk_context_t* ctx) {
   }
   #else
   kk_with_string_as_qutf8_borrow(cmd, ccmd, ctx) {
-    exitcode = system(ccmd);
+    int status = system(ccmd);
+    if (WIFEXITED(status)) {
+      exitcode = WEXITSTATUS(status);
+    } else if (WIFSIGNALED(status)) { 
+      exitcode = -WTERMSIG(status); // Like Haskell, return negative code for signal termination
+    } else {
+      // Technically WIFSTOPPED(status) can happen, but only if the process is being traced, or the call was done with WUNTRACED.
+      kk_fatal_error(EINVAL, "kk_os_run_system: unexpected termination");
+    }
   }
   #endif
   kk_string_drop(cmd, ctx);


timbertson:
I'm surprised `stopped` is a different branch, isn't that just another signal?

FWIW in my WIP libuv process branch, I went with [an ADT for Exit-status(int) and Exit-signal(int)](https://github.com/timbertson/koka/blob/012b49c2382e59c98522b7424ad10c0ffdc062a8/lib/std/async/process.kk#L69-L83). I don't expect most people would need to inspect this directly, they just want to check for success or not.

**Edit**: Ah, looks like `stopped` would not happen here, according to [the docs](https://man7.org/linux/man-pages/man2/wait.2.html):

> WIFSTOPPED(wstatus)
              returns true if the child process was stopped by delivery
              of a signal; this is possible only if the call was done
              using WUNTRACED or when the child is being traced (see
              [ptrace(2)](https://man7.org/linux/man-pages/man2/ptrace.2.html)).

kklib/src/os.c
```diff
@@ -648,7 +648,14 @@ kk_decl_export int kk_os_run_system(kk_string_t cmd, kk_context_t* ctx) {
   }
   #else
   kk_with_string_as_qutf8_borrow(cmd, ccmd, ctx) {
-    exitcode = system(ccmd);
+    int status = system(ccmd);
+    if (WIFEXITED(status)) {
+      exitcode = WEXITSTATUS(status);
+    } else if (WIFSIGNALED(status)) { // TODO: Return differently in this branch
+      exitcode = WTERMSIG(status);
```
timbertson: FWIW haskell returns the negative of this value to distinguish it from an exit status, based on my various failed runs which return `-11` (SIGSEGV) 🤦

timbertson (approved):

PR #804, opened 2025-09-12
Fix open constructor tags #711

### LINKED ISSUE #711: Extending exn-info fails C compilation if the module is explicitly named
It seems unlikely I've captured the real reason, because I've seen other modules successfully extend `exn-info`. But I've seen this error a number of times, and this is the simplest way to reproduce that I found.

```
// extend-test.kk
// module extend-test
extend type exception-info
  ExtraInfo(s: string)

fun main() { () }
```

That compiles and runs, but if I uncomment the `module extend-test` line, it fails with:

```
check   : extend-test
.koka/v3.1.3/clang-debug-23345a/extend_dash_test.c:3:35: error: static declaration of 'kk_extend_dash_test__tag_ExtraInfo' follows non-static declaration
kk_declare_string_literal(static, kk_extend_dash_test__tag_ExtraInfo, 21, "extend-test/ExtraInfo")
                                  ^
.koka/v3.1.3/clang-debug-23345a/extend_dash_test.h:38:20: note: previous declaration is here
extern kk_string_t kk_extend_dash_test__tag_ExtraInfo;
                   ^
1 error generated.
extend-test(1, 1): build error: user error (error  : command failed (exit code 1)
command: /usr/bin/clang -Wall -Wextra -Wpointer-arith -Wshadow -Wstrict-aliasing -Wno-unknown-pragmas -Wno-missing-field-initializers -Wno-unused-parameter -Wno-unused-variable -Wno-unused-value -Wno-unused-but-set-variable -Wno-cast-qual -Wno-undef -Wno-reserved-id-macro -Wno-unused-macros -Wno-cast-align -Wno-unknown-warning-option -g -Og -c -I /Users/tcuthbertson/.local/share/koka/v3.1.3/kklib/include -DKK_MIMALLOC=8 -o .koka/v3.1.3/clang-debug-23345a/extend_dash_test.o .koka/v3.1.3/clang-debug-23345a/extend_dash_test.c)

Failed to compile extend-test.kk
```

Koka version dev (8d9004e9), although I'm pretty sure I've seen this same error on previous versions too.

TimWhiting:
Fixes #711. 

The real fix would be to move all private constructor related functions out of the C header files, but currently we have them as `static inline` functions in the headers. I attempted to move all of those functions depend on the visibility of constructor, but ran into an issue where we would have to rewrite a ton of the standard library .c/.h files which use this pattern in their headers. Also moving them to C files prevents the C compiler from doing as much cross-compilation-unit optimization - which is likely crucial for some private constructors like `Cctx`, and `Ev`.

### DIFF
--- a/src/Kind/Infer.hs
+++ b/src/Kind/Infer.hs
@@ -312,7 +312,8 @@ synConstrTag (con)
   = let name = toOpenTagName (unqualify (conInfoName con))
         rc   = rangeHide (conInfoRange con)
         expr = Lit (LitString (show (conInfoName con)) rc)
-    in DefNonRec (Def (ValueBinder name () expr rc rc) rc (conInfoVis con) DefVal InlineNever "")
+        -- Tags must be public, since they may extend a type outside of the current module
+    in DefNonRec (Def (ValueBinder name () expr rc rc) rc Public DefVal InlineNever "") 
 
 {- ------------------------------------------------------------------------------------------------------------
    Lazy types and constructors


daanx:
Apologies for the late merge! Thanks!

PR #816, opened 2025-10-07
Fix incorrect comparison implementations

chtenb:
It took me a while to figure out why my algorithm behaved so strange 😅


TimWhiting:
This only affects the equality case right? The order of arguments were swapped before so the actual ordering would have been good.

chtenb:
yes

PR #831, opened 2025-11-02
Stop re-allocating static strings for every local string definition

omegametabroccolo:
Tracing currently allocates memory that isn't freed:

```koka
fun main()
  while { True }
    trace("ping")
```

This fix was suggested by @TimWhiting.

### DIFF
--- a/src/Backend/C/FromCore.hs
+++ b/src/Backend/C/FromCore.hs
@@ -1759,7 +1759,7 @@ genExprPrim expr
              if (s=="")
               then return ([],text "kk_string_empty()")
               else do let (cstr,clen) = cstring s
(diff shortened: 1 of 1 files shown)


TimWhiting:
Note this is not specific to trace, it re-allocates static strings for every local string definition (I would change the commit message and title of the PR). I don't know if this solution is 100% correct though for every situation. I'd want to dig deeper to make sure that it really does fix the issue, and doesn't create new ones.

daanx:
ah interesting bug! Since strings are immutable, it should indeed be fine to declare it as static; it is important though that the actual string is initially allocated and copied from the static data into the heap as we might use a compressed heap (and all pointers must be reachable within some offset of the base)

daanx:
Thanks!

PR #835, opened 2025-11-12
Fix `--fstdalloc` on OpenBSD

omegametabroccolo:
OpenBSD removed `malloc_usable_size` from libc [in version 6.5](https://www.openbsd.org/plus65.html), we need to keep track of each allocation's size ourselves. Of course doing this adds some memory overhead, especially when doing lots of small allocations, but I think it's better than nothing. The test suite runs fine on my OpenBSD -current machine when adding `--fstdalloc` to `commonFlags` in `test/Spec.hs`.

### DIFF
--- a/kklib/include/kklib.h
+++ b/kklib/include/kklib.h
@@ -534,6 +534,49 @@ static inline void kk_free_local(const void* p, kk_context_t* ctx) {
 
 #define kk_malloc_usable_size(p)  mi_usable_size(p)
 
+#elif defined(__OpenBSD__)
+
+// On OpenBSD, libc's malloc implementation doesn't allow checking the
+// size of an allocation. Here we prefix each allocation with a `size_t`
+// that contains its size (excluding the prefix itself, so that
+// `kk_malloc_usable_size` works the same as on other platforms).
+
+static inline void* kk_malloc(kk_ssize_t sz, kk_context_t* ctx) {
+  kk_unused(ctx);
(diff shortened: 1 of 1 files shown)


daanx:
Thanks! .. but yikes, that can cause a lot of overhead just to support constructor contexts (which are the only feature that need to copy arbitrary blocks). As an aside, do you need to use the standard allocator on FreeBSD?

omegametabroccolo:
> As an aside, do you need to use the standard allocator on FreeBSD?

I don't know about FreeBSD, but assuming you meant OpenBSD, `mimalloc` works just fine, I'd just like to be able to use the system's allocator which comes with a bunch of security features.

> constructor contexts (which are the only feature that need to copy arbitrary blocks)

Maybe we could make constructor contexts carry their own size so that all other allocations don't have to. That seems a bit more involved than this PR though, especially if we don't want to have an impact on systems that do have a way to get an allocation's size. I'll try to look into it.

daanx:
Ah yes, OpenBSD. We can't really have just constructor contexts carry their size as they can switch between the actual type (say list<a>) and their context. We could perhaps optimize a bit: for any regular type that does not have "raw" fields (like `int64` etc) we can determine the size from the scan filed in the header. Only for types with raw fields (that follow the scan fields) we need to have the size. However, perhaps there is some clever thing we can do here.. but not sure. Also, perhaps a bit too much just to support OpenBSD -- the current PR solves it as well :-)  I'll ponder it a bit more.

PR #839, opened 2025-11-25
Fix #782

### LINKED ISSUE #782: `1.1.l(1.1).l(1.1)` causes compiler internal error
The following code triggers a compiler internal error:

```koka
import std/num/float64

fun l(x : float64, y : float64)
  x + y

fun main()
  1.1.l(1.1).l(1.1)
```

Error message:

```
main/@main(1, 1): internal error: Core.Simplify.topDownLet.liftLets
CallStack (from HasCallStack):
  error, called at src\Common\Failure.hs:46:12 in koka-3.2.2-...
  raise, called at src\Common\Failure.hs:32:5 in koka-3.2.2-...
  failure, called at src\Common\Failure.hs:28:11 in koka-3.2.2-...
  assertion, called at src\Core\Simplify.hs:109:9 in koka-3.2.2-...
  topDown, called at src\Core\Simplify.hs:674:16 in koka-3.2.2-...
```

If only one `.l(1.1)` call is made, the code compiles without problems, for example:

```koka
import std/num/float64

fun l(x : float64, y : float64)
  x + y

fun main()
  1.1.l(1.1)
```

TimWhiting:
Inlining a function multiple times caused shadowing of variables without uniquefying. Instead of missing the opportunity or crashing, just uniquefy when needed.

@daanx This passes the tests, but I'm not as familiar with this code.


daanx:
Thanks!

PR #849, opened 2025-12-08
Add `prepend` to the `exn` module

omegametabroccolo:
I find myself using `prepend` a lot when writing system call bindings. It's currently defined in three different places in the standard library, I think it makes sense to export it from `exn` instead.


daanx:
Thanks so much!

PR #852, opened 2025-12-16
LibUV / Async Basics

TimWhiting:
Here is a more simple PR, without files, streams, etc, and cleaned up and documented based on things I've learned since I implemented it last time.

This PR contributes a working async event loop + timers on C/JS/WASM. I've tested with a leak checker on MacOS to ensure no memory leaks.


lib/std/async/null.kk
```diff
@@ -0,0 +1,38 @@
+// Used for external interfaces
+module std/async/null
```
timbertson, line 2: Not async specific, maybe this belongs in something like std/ffi?

lib/uv/inline/timer.c
```diff
@@ -0,0 +1,225 @@
...
+  uv_timer_t* uv_timer = (uv_timer_t*)kk_tm_to_uv(timer);
+  uv_timer_stop(uv_timer);
+  return kk_Unit;
+}
+
+// Actually clean up the timer 
+// This drops the callback first in case it is holding onto the timer - as it does for uv/timer/timer()
+kk_unit_t kk_libuv_timer_finish(kk_uv_timer__timer timer, kk_context_t* _ctx) {
+  kk_timer_t* kk_timer = kk_tm_to_uv(timer);
+  if (kk_likely(!kk_function_is_null(kk_timer->callback, kk_context()))) {
```
timbertson: Should kk_libuv_timer_init set this to null in the first place?

lib/uv/inline/timer.c
```diff
@@ -0,0 +1,225 @@
...
+// The uv callback for the timer
+void kk_uv_timer_unit_callback(uv_timer_t* uv_timer) {
+  kk_context_t* _ctx = kk_get_context();
+  kk_timer_t* kk_timer = (kk_timer_t*)uv_timer;
+  kk_function_t callback = kk_timer->callback; // Get the callback
+  if (uv_timer_get_repeat(uv_timer) == 0) { // If this is a one-shot timer, just call the callback
+    kk_unit_callback(callback, kk_context());
+    return;
+  } else { // Otherwise, we need to dup the callback, as it will be called again
+    kk_function_dup(callback, kk_context());
```
timbertson:
```suggestion
    callback = kk_function_dup(callback, kk_context());
```

(I don't know if it ever matters in practice)

lib/std/async/null.kk
TimWhiting: Good question. I'll let Daan decide the namespace for this.

lib/uv/inline/event-loop.c
```diff
@@ -0,0 +1,123 @@
...
+  int ret = uv_run(uvloop(), UV_RUN_DEFAULT);
+  if (ret != 0){
+    kk_warning_message("Event loop closed with status %s\n", uv_err_name(ret));
+  }
+}
+
+static void kk_uv_loop_close(kk_context_t* _ctx) {
+  int ret = uv_loop_close(uvloop());
+  if (ret != 0) {
+    kk_warning_message("Event loop closed %s\n", uv_err_name(ret));
```
timbertson: IMO this should be an error that we propagate, and if it reaches the default handler then that can either throw (if allowed) or exit with nonzero status.

Usually a failure in uv_loop_close indicates an improperly cleaned up handle, which is worth knowing about.

TimWhiting: Yeah, I considered making it just exit immediately actually. I'm not sure what the purpose of letting the program handle this would be, since it almost surely means that there is a serious error in the program or in our libUV bindings. Ideally we would enforce that all handles are cleaned up (references don't escape the scope of the loop) via a scoped type parameter like the local or state effect `st<h>`. I'm not sure that is the best idea though, locals are typically locally guaranteed to not escape, and `st<h>` often is used with global or essentially for pretty local state. Having such a type parameter go throughout the program might be unwieldy. 

Maybe a compromise would be to add a scoped type parameter, and then expose the resources (or just apis to access parts of the resources) to the rest of the code via effect handlers. Maybe named handlers actually. I'm curious how timers, etc would look with this design.

Issue #537, opened 2024-05-24
Struct Constructor TitleCase


TimWhiting:
Struct constructors should really be converted from `type-case` to `ConstructorCase`.

Most type definitions use `ConstructorCase`, and to have the autogenerated constructors for structs be different is quite a bit weird.

PR #859, opened 2026-01-07
Make `struct` constructors PascalCase

omegametabroccolo:
This PR implements #537 and fixes uses of `struct` constructors in `lib/std` and `test`. Not everything was covered by tests so I had to inspect the output of `grep -R '[A-Z][a-z]*-[a-z]+'` manually, I hope I didn't miss anything. Not covered:
- `type` constructors that don't follow this convention
- `lib/v1`


TimWhiting:
@daanx 

I think this is a useful and important change even though it breaks current code. 
It matches Koka's style for many existing multiword `type` constructors. 
In my code I've made all `struct`s `type`s for now and forced this convention.

It is better to make this breaking change now while the community is small and almost entirely on the community discord (judging by the Koka code I've seen on github and the number of users in the discord).

PR #870, opened 2026-03-11
Add profiling support (--fprofile) for gcc and clang

angelica-moreira:
Adds --fprofile flag for profiling Koka programs with gcc (gprof) and clang (llvm instrumentation).

Changes:
- --fprofile flag: uses -pg/gprof for gcc, -fprofile-instr-generate for clang
- cold path annotations on RC slow-path functions in kklib
- multiarch library search path fix for Debian/Ubuntu (/usr/lib/x86_64-linux-gnu)
- profiling utility (util/profile.kk) with --help flag
- readme section covering gcc, clang, and linux perf profiling


daanx:
Thanks Angelica, this is great to have!

Issue #872, opened 2026-03-21
Capture-avoiding substitution for forall types


TimWhiting:
## Summary

The `substitute` function for `TForall` in `src/Type/TypeVar.hs` (lines 520-539) has an assertion that fires when bound variables in a forall capture free variables from the substitution range. There is commented-out code (from commit a30efb42c) that attempted to fix this by alpha-renaming bound variables with a `+2000` ID offset, but it was disabled because the offset IDs could collide with meta variables from the unique counter during type checking.

## Proposed fix

Use **negative IDs** for alpha-renamed bound variables during capture-avoidance. Since the unique counter only produces positive IDs, negative IDs are guaranteed to never collide with any compiler-generated variable. For nesting safety, compute the minimum ID in scope and go below it.

## Concerns

- Negative type variable IDs could appear during inference and might cause issues in code that assumes non-negative IDs
- The `tvsMax` function (used in 3 places) returns 0 for all-negative sets, which might produce unexpected behavior
- Pretty printing would show negative numbers for capture-renamed bound variables

## Current status

Not currently needed — the canonical kind-based ordering fix (PR #873) eliminates the specific ID collisions that triggered this assertion for the `test/cgen/type-of.kk` test case. However, the underlying variable capture issue could theoretically occur in other scenarios.

## Related

- Commit a30efb42c: original attempt at capture-avoidance (commented out)
- TypeVar.hs line 516: the assertion that fires on capture

PR #873, opened 2026-03-21
Fix typeOf failures with higher-rank types in effect handlers

TimWhiting:
## Summary

- Fixes `typeOf` crash (`KindMismatch` assertion in `subNew`) when compiling recursive functions with higher-rank effect handler types
- Root cause: forall type variable ordering could differ between the assumed type (from annotations) and the final generalized type, causing `TypeApp` arguments to be misaligned with the `Var`'s forall variable order

## Changes

**Canonical forall variable ordering** — Sort generalized type variables by kind group during generalization:
- `Kind.hs`: Add `compareKindCanonical` — non-special kinds (value types) sort before special kinds (effects, heaps, labels), matching the existing `oftv` convention. Within each group, all kinds compare as `EQ` to preserve `ofuv` occurrence order (stable sort).
- `InferMonad.hs`: Apply canonical sort in `generalizeX` when collecting type variables for generalization
- `Type.hs`: Add `canonicalizeScheme` to sort an existing type scheme's forall variables into canonical order
- `Infer.hs`: Canonicalize annotation types in `createGammas` so assumed types for recursive definitions have the same forall order as the final generalized types

**Test case** — `test/cgen/type-of.kk`: exercises higher-rank forall types with effect handlers and recursive calls

## Why the ordering matters

When type-checking recursive definitions, the compiler:
1. Creates an assumed type for the function (from annotations, generalized during kind inference)
2. Type-checks the body, creating `TypeApp` nodes for recursive calls with arguments in the assumed type's forall order
3. Re-generalizes the function, potentially producing a different forall variable order (because `ofuv` occurrence order depends on type structure, which changes after unification)
4. Replaces `Var` references with the final type via `|~>`, but `TypeApp` arguments keep the old order

The canonical sort ensures steps 1 and 3 produce the same ordering, eliminating the mismatch.

## Related

- #872: potential future improvement for capture-avoiding substitution in `TypeVar.hs`

## Test plan

- [x] `stack run koka -- -e test/cgen/type-of.kk` prints "runs"
- [x] `stack test --fast` — 407 examples, 0 new failures (2 pre-existing unicode locale failures)


TimWhiting:
Superseded by new PR with simpler fix (replaceRecCallEx only, no canonical sort)

PR #877, opened 2026-03-24
Fix prefix/postfix identifiers

TimWhiting:
@daanx 

This fixes a lot of the .kki format errors, and I've included a test file that covers a lot of the common usages of make unique symbolic names from the code.


TimWhiting:
I'm looking into another .kki issue, that might have been caused by using this PR. I'll mark as non-draft when ready.

daanx:
Hi Tim -- thanks for this! I revised the code manually based on your PR -- hope it works well now.

PR #885, opened 2026-03-27
fix rangeMapFindIn containment filter

cuiweixie:
## Summary

In `rangeMapFindIn`, the filter used `||`, so a range satisfied the predicate if either edge was inside the query span, which matches far too many entries. Use `&&` so only ranges **fully contained** in the query range are kept (consistent with needing both `rangeStart >= start` and `rangeEnd <= end`).


daanx:
Thanks for the PR. Can you give an example where I can see the difference? Is this onHover inside vs code?

PR #905, opened 2026-07-11
inference: fix implicit-evidence placement (unbound iev variables in generated C)

TimWhiting:
Solved implicit-constraint evidence (e.g. the `hdiv` evidence `iev@N` variables) is let-bound around whatever core is under generalization when the solver runs. When the constraint's occurrences live in a *sibling* scope — e.g. lambdas next to a local named `fun` (whose `inferIsolated` generalization solves the pending constraints) — the evidence def lands as a dead let in the sibling while the occurrences stay unbound. This is invisible while the evidence is erased (`local-get`'s inline extern drops its evidence argument), but the monadic transform lifts the dangling variables into yield continuations as real parameters, producing uncompilable C:

```
error: use of undeclared identifier 'iev_424'
  ... kk_std_core_hnd_yield_extend(new_lis_fun555(iev_424, iev_492, ...), _ctx);
```

Fix: decouple solving from placement. Solved defs queue in inference state (`pendingEvDefs`); `placeEvidenceDefs` at each generalization binds exactly those whose evidence variable occurs free in that core; defs with no occurrence anywhere (speculative elaboration) are dropped at `zapSubst` — sound, since evidence values are pure closed constructors.

Regression test: `test/cgen/iev-evidence-placement.kk` (fails on dev with 15 undeclared-`iev_N` errors; compiles and prints `3,4,5,9` with the fix). Full suite passes.


TimWhiting:
Closing to rethink: this queues solved evidence and places it by occurrence, which patches the symptom (defs attached to a sibling's core) rather than fixing why a sibling's generalization drains constraints that aren't in its scope. Will follow up with a scoping-level fix.

PR #906, opened 2026-07-11
specialize: terminate mutual-recursion specialization (ancestors only at smaller instantiated types)

TimWhiting:
The specializer's only termination guard is "don't specialize yourself" (`inlineName /= thisDefName`). Mutually recursive specializable functions — each with a function-typed argument, e.g. `build -> patch -> patch-kids -> build ...` with a callback — walk around it: each member's specialized body exposes a specializable call to the next, minting a fresh `@spec` name every round (the lambda argument is a fresh copy each time, so no cache applies). `-O2` builds of such code never terminate.

Fix: only respecialize an **ancestor** at a **strictly smaller instantiation weight** — the constructor count of the call's type arguments, the same decreasing measure implicit inference uses (`isDecreasingChain` in `Type/InferMonad.hs`). `SpecM` tracks the minimum ancestral weight per name; the weight must strictly decrease, so specialization terminates. Multi-step specialization at shrinking types (`twostep-large2`'s `repeatN` at `list<list<int>>`, `list<int>`, `int`) is unaffected; same-type cycles terminate after one round. `specOneExpr`'s traversal is hand-rolled so the ancestry scopes over the descent into freshly created specialization bodies.

New test: `test/cgen/specialize/mutualspec{,-lib}.kk` — it makes the specializer diverge on current `dev` (at both `-O1` and `-O2`) and passes with this fix. All `cgen/specialize` tests pass; full suite passes.


TimWhiting:
Addressed the review: the `ioc` effect annotations were left over from the originating codebase and don't exist on `dev` — the library module now uses plain `<div,exn>` and typechecks, and the `.out` is regenerated from the real signature dump. Notably, with the effects fixed the test now genuinely reproduces the divergence on current `dev` (hangs at -O1 and -O2 without this fix). Also kept `UniqueT` abstract by adding `hoistUniqueT` to `Common.Unique` instead of exporting the constructor.

PR #909, opened 2026-07-12
compile record copy (m(field = e)) to match+reconstruct, enabling in-place update

TimWhiting:
The record-update idiom `m(field = e)` previously compiled (even at -O2, with the copy inlined) to one runtime `OptArg`/`NoOptArg` scrutiny per field plus one borrowing accessor match per unchanged field — and on the unique path Perceus `free`d the record and allocated a fresh one.

It now compiles to a single destructure that Perceus turns into an in-place field update when the value is unique (falling back to dup+alloc when shared):

```
fun bump(m) m(a = m.a + 1)
-- core at -O2:
match m { Model(a,b,c,d,e) -> Model(a+1, b, c, d, e) }   -- with constructor reuse
```

Three changes:

- **Copy synthesis** (`Kind/Infer.synCopyCon`): the copy constructor is one match on `@this` reconstructing with either the given optional argument or the matched field, instead of per-field accessor calls as optional-argument defaults. No call-site-specialized copies: the definition inlines (`InlineAlways`) and the optional arguments — always literal `OptArg`/`NoOptArg` at a call site — fold away statically.

- **Simplify, known-constructor folding** (`kmatchPattern`): answers `NoMatch` on constructor-name mismatch regardless of pattern arity (a single `Unknown` previously blocked branch selection entirely), handles nullary polymorphic constructor scrutinees and `@box` applications against `@Box` patterns, and A-normalizes non-total constructor arguments in let bindings (`val x = Con(f(e))` ⇒ `val y = f(e); val x = Con(y)`, same evaluation order) so the constructor binding is total and can inline into its match when used once (splitting only non-total arguments makes this a one-step fixpoint — total ones would inline back).

- **Simplify, known scrutinee**: trailing let bindings sink into an irrefutable singleton match, and a match on a scrutinee variable folds inside a branch that already destructured it (wildcard fields are enriched to fresh pattern variables as needed) — so the `m.a` read in `m(a = m.a + 1)` becomes the pattern variable itself instead of a second borrow-match.

10M single-field updates of a five-field struct: **62ms → 16ms**, flat rss.

`test/parc/parc22`/`parc23` recorded core is regenerated and now pins the reuse shape (`@con-fields-assign` on the unique path inside `@copy`); `parc15`/`parc16` exercise nested same-scrutinee matches, which now fold before Parc, and are regenerated likewise. Full test suite passes.

### DIFF
--- a/src/Core/Simplify.hs
+++ b/src/Core/Simplify.hs
@@ -606,7 +606,7 @@ kmatchPattern scrut@(Lit lit) (PatLit pLit)
 
 kmatchPattern scrut@(Con name _repr) (PatCon pname pats _prepr _ _ _ _info _)
   = --trace ("kmatchPat PatCon empty pats " ++ show name ++ " ___pat___ " ++ show pname) $
-    if name /= pname then NoMatch    -- different constructor: no match, whatever its arity
+    if name /= pname then NoMatch    -- different constructor: no match
     else if null pats then Match ([], scrut)
     else Unknown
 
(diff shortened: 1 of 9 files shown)


TimWhiting:
Soundness note on the known-scrutinee fold (folding a nested match on an already-destructured variable): it relies on the scrutinee *binding* being immutable between the two matches, which holds in Koka core with one exception.

- **ref / var contents**: safe by construction — the fold only unifies matches on the same core variable, and every `!r` / `local-get` read is an application producing a fresh binding, so matches on two reads never unify. Interior `ref` fields bind the ref pointer (immutable identity), not its contents.
- **Perceus reuse**: runs after Parc; the fold runs on pre-Parc, immutable core.
- **Lazy values**: the real exception — forcing overwrites the cell in place, so a match before a force and a match after can see different constructors. Thunk-constructor patterns are now excluded as witnesses (and from irrefutability for the let-sinking rule); whnf-constructor patterns remain eligible since forcing is monotone (thunk → value, never back).

